PR #69, opened 2015-07-02
Dev

jjchico:
Dear sandain,
I find mscs incredibly useful. Thanks for the great job. I made some fixing with /etc/default/mscs sourcing. I hope you find them useful.

Best.


jjchico:
Also fixed a problem with mscs list-backups not listing current (last) backup, only previous increments.

PR #72, opened 2015-07-04
Fix /etc/default/mscs loading

jjchico:
/etc/default/mscs was sourced 'after' default variables were used
making overrides not to work. /etc/default/mscs is now sourced twice,
after LOCATION definition and after other variables definition so all
variables can be overrided. A better fix is let for later.


sandain:
This change was merged with commit e0ee185be9d2d1c8f143ea734e326b64d463ff53.  Removing duplicate pull request.

PR #74, opened 2015-07-07
Gitignores for ctags

jjchico:
I find it very useful to use ctags and/or ctags-aware editors to navigate through the code, so I added some ctags' gitignores. I hope you find them useful too.


sandain:
Looks good. Thanks.

PR #75, opened 2015-07-07
Better world lists

jjchico:
I always check used ports in the config files when adding a new world, so I changed "mscs ls" to print the used port of every server/world.

'ls' was not documented in 'mscs help' so changed that too.

I modified 'list' option to generate a more detailed list including some server.properties parameters. The modifications were easy thanks to the rich set of procedures already implemented in the script. The exact list of properties shown is configured through a global variable. It is intended to give the user a general view of the configured servers and major options in use.

Now, a wrong <option> in 'mscs list <option>' will abort with an error message instead of listing enabled worlds. No option will still print an enabled world's listing.

There are a couple of changes that I also want to make in the listings in case you find them right for master:

1) Change 'mscs ls|list' behaviour with no option to list both enabled and disabled worlds.
2) Include the enabled/disabled state of the world in the listing.

I hope you find these changes ready for integration. If not, please let me know.

### DIFF
--- a/mscs
+++ b/mscs
@@ -83,13 +83,17 @@ Options:
   enable <world>
     Enable a disabled world server.
 
-  list <option>
+  ls <option>
     Display a list of worlds.
     Options:
       enabled   Display a list of enabled worlds, default.
       disabled  Display a list of disabled worlds.
       running   Display a list of running worlds.
       stopped   Display a list of stopped worlds.
+    If no option, 'enabled' is assumed.
+
+  list <option>
+    Same as 'ls' but more detailed.
 
   status <world>
     Display the status of the Minecraft world server.  Display the status of
@@ -280,6 +284,13 @@ BACKUP_DURATION=15
 
 # Length in days that logs survive.
 LOG_DURATION=15
+
+
+# Listing options
+# ---------------------------------------------------------------------------
+
+# Server properties for detailed listing (list).
+DETAILED_LISTING_PROPERTIES="motd server-ip server-port max-players level-type online-mode"
 
 
 # Mirror Image Options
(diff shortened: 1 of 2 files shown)


sandain:
I'd like to add that I like the global variable and the way that you implemented the change.  I just don't want to accept this patch until we work out the details.

jjchico:
I realized the problem with mscs.completion later. It can be solved by parsing the 'mscs ls' output with sed inside mscs.completion:

    sed -n '/:/{s/^ *//;s/:.*$//;p}'

Basically, sed removes leading spaces and trailing characters starting at ':'. This way, 'ls' can be informative for humans and still useful for completion. It would also work with the detailed list generated by the new version of 'list'. Of course, a new listing mode could be implemented just for completion parsing, but I think that 'ls' would do.

I also thought about adding to status/show but these imply to query the server so I think it is better to have list/ls to show configuration and status/show to get actual information from the running process.

What about the additional proposed change: list both enabled and disabled worlds by default? I really find it useful to avoid having to issue two 'list' commands to see the whole set of available servers.

jjchico:
So I finally made the changes to list all worlds by default and inform about the enable/disable state. I added two new methods: isWorldEnabled() and getAvailableWorlds() to simplify things. I tried to be consistent with current code.

jjchico:
mscs.completion is fixed.
In the meantime, I discovered some other "bugs" like that a world cannot be deleted if it is not enabled! It is related to how enabled/disabled worlds are handled by the script but I better open a bug/improvement request so that we can discuss it there.

sandain:
I'm going to hold off on merging this until we have some consensus on the enable/disable world option since you are touching that code.

Also, if your changes to the `mscs.completion` script will work with either long or short form list, is it really necessary to have both options?  Why not just leave `ls` and `list` as synonyms?

jjchico:
OK with waiting for a final decision on enable/disable handling.

I find both 'ls' and 'list' useful: I commonly use 'ls' to quickly check the ports I have already used when I am configuring a new client or I am creating a new server. I like to have just the essential information for this. But sometimes you have to go to server.properties to check some additional options. Here 'list' is very useful to have all the interesting details (configurable) that you may need from time to time without having to open all the server.properties files.

PR #83, opened 2015-07-29
Fix restore

jjchico:
Add "--force" to rdiff-backup so that backups can be restored.
Refuse to restore if server is running.
Fix date/time listing in list-backups.

### DIFF
--- a/mscs
+++ b/mscs
@@ -2239,16 +2239,20 @@ case "$1" in
     fi
   ;;
   restore-backup)
-    if [ -n "$2" ] && [ $(listContains "$2" "$(getEnabledWorlds)") -eq 1 ]; then
-      worldBackupRestore "$2" "$3"
-    elif [ -n "$2" ]; then
-      printf "World '$2' not recognized.\n"
-      printf "  Usage:  $0 $1 <world>\n"
-      exit 1
-    else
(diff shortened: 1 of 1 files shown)


zanix:
Nice, thanks for taking a look at this.
I'm almost thinking that the contents of the world should be removed before the restore because if any new files were created since the backup was made such as new regions or chunks, those files will still exist.

jjchico:
You are right. Even better, restoration should happen in a temporary directory and only if the restoration was OK should it be moved to the original location. I will look into this.

zanix:
That does sound better

jjchico:
In order to fix it properly I would like to migrate the return value of some methods (like isWorld*) to proper return values instead of using stdout. I would also add sensible return values for other methods to make further development easier (like in worldBackup) so that you can do something like:

    if [ worldBackup $world ]; then
      echo "Success!"
    fi

Since it means some changes on the coding style, would it be acceptable as a PR?

sandain:
Thanks for the fix!  I probably would have just initiated a stop sequence rather than tossing out an error message, but that works too.

jjchico:
Thanks @sandain. This merge introduces some conflicts with my other PR. I will do a merge in the other branch.

PR #84, opened 2015-07-29
Change method return values to real ret values not stdout

jjchico:
* isWorldAvailable, runningWorld, etc.
* less code
* faster code
* easier code

### DIFF
--- a/mscs
+++ b/mscs
@@ -359,7 +359,7 @@ execute() {
 # Get the PID of the Java process for the world server.
 #
 # @param 1 The world server of interest.
-# @return The Java PID.
+# @stdout The Java PID.
 # ---------------------------------------------------------------------------
 getJavaPID() {
   local PID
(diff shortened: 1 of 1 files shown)


jjchico:
Lots of questions :)

  * I missed the Doxygen thing so it should be reverted to @return. I am sorry.
  * "[" is just a short-cut for command "test", although most shells have an internal implementation. "[" just evaluate a list of conditions and returs "0" if true or "1" if false (not kidding) just like any other command. The argument to "if" is actually the output of any commando, not just "test" or "[". Yes it all works with dash: it is ancient bourne shell syntax.
  * listContains() currently returns true if both the item and the list are "". The -n "" check can be included but it may not be necessary since most checks now use isWorldEnabled() or isWorldAvailable() which already exclude this (or should). I have planned to check it soon.
  * The impact of calling serverRunning twice in the overview() method should be acceptable, but for consistency and robustness it should be called once and use the recorded value throughout the method. We do not want to run just half of the method because the server crashed in bad moment.
  * Backup/restore works without the permission changes (as it should). Changing the permissions in the script may lead to a security/privacy risk. I do not really know how to get this particular change out of the branch's history. Does a rebase allow for particular patch edits inside a commit? I may find out if you insist :)

jjchico:
I re-did the changes so that they apply to master more cleanly (I hope). Making a new PR in a minute.

PR #86, opened 2015-08-01
Ret val

jjchico:
Change return value style of some methods to simplify code and improve readability and performance. Also add return value to worldBackup for error handling.


sandain:
Thanks for the patches.  I'll go ahead and accept these patches, but I would have preferred that patch 305ab21a805b8a397691748cdc447d9eadd0cfae was merged into 894569976353aba70807b86256c2526ee395202f to avoid the needless code churn.

Issue #82, opened 2015-07-28
Backup restore fails


zanix:
mouarflenoob from the Minecraft Forum noticed a possible bug in the backup restore process
```
mscs restore-backup alpha 2015-07-25T06:50:51+02:00
Fatal Error: Restore target /opt/mscs/worlds/alpha already exists, specify --force to overwrite.
```

This can be caused by a number of reasons:

`rdiff-backup-data` directory exists in restore location:
rdiff-backup keeps its administration on the mirror in `<backup directory>/rdiff-backup-data`.
If you accidently copy `rdiff-backup-data` to the source directory, rdiff-backup will refuse to backup.
This makes sense as it would overwrite the administration directory at the backup location.
Unfortunately the error message it gives is rather misleading:
```
Fatal Error: Restore target /opt/mscs/worlds/alpha already exists, specify --force to overwrite.
```

The source directory already exists.
I haven't had time to double check this, but I suspect that rdiff-backup will not restore files if the destination already exists. In my original documentation on the wiki, I never specified having to remove or rename the restore location and I haven't had to perform a restore using rdiff-backup lately.

I think to solve this, the script could run a check on the user permissions of the backup directory, check for a directory named `rdiff-backup-data` in the worlds location, rename or remove the restore location, and/or add some more documentation to the wiki.

jjchico:
Using --force is the right option here because mscs will do a backup before any restore. I already did this [fix](https://github.com/jjchico/MinecraftServerControlScript/tree/fix-restore).

But the main issue is that the restore is done without checking if the previous backup was right and without checking if the server is running. I would just refuse to restore and tell the user to stop the server first manually.

I'll add these changes when I have a minute.

jjchico:
It should be fixed by now, except for the previous backup checking. I'll make a PR.

PR #88, opened 2015-08-03
Smarter and safer worldBackupRestore

jjchico:
* Backup prior to restore is done outside worldBackupRestore
    to improve reusability of the method.
  * Restoration is cancelled if the world cannot be backed-up first
  * Restoration is done in a tmp dir first and world is overwritten
    only if successful
  * Fixes issue #82


zanix:
Looks good. Great idea to restore to a temp location first.

Issue #65, opened 2014-12-27
invalid number error


MattPark:
So for me, server status isn't working.  All the commands that invoke status in some way, run into an issue with "invalid number".

It's because 08 and 09 are interpreted as invalid octal numbers, and when you try to printf them into your status commands, it causes a failure.

I tried to force it to cast them as decimal by evaluating as $(( 10#09 )) but it's not working.

I also tried switching the shells from /bin/sh to /bin/bash but that creates a lot more errors.  Not sure how best to resolve this

MattPark:
Short answer no, it doesn't seem to do it.
```
[~]
[root@nas]# /opt/bin/bash -c "echo $(printf "%d" $(echo 99))"
99
[~]
[root@nas]# /opt/bin/bash -c "echo $(printf "%d" $(echo 09))"
bash: printf: 09: invalid number
0
[~]
[root@nas]# /opt/bin/bash -c "echo $(printf "%d" $(echo 07))"
7
[~]
[root@nas]# /opt/bin/bash -c "echo $(printf "%d" $(echo 08))"
bash: printf: 08: invalid number
0
[~]
[root@nas]#
```

sandain:
I'm not totaly sure what I'm seeing here.  I would expect all of those to fail or succeed in the same way, especially those that have a leading zero.

This may be related to the way you are doing the quoting: the `%d` is probably not being quoted when it reaches the printf statement.  Try swapping the outer double quotes with single quotes for a better test:
```
/opt/bin/bash -c 'echo $(printf "%d" $(echo 99))'
/opt/bin/bash -c 'echo $(printf "%d" $(echo 09))'
/opt/bin/bash -c 'echo $(printf "%d" $(echo 08))'
/opt/bin/bash -c 'echo $(printf "%d" $(echo 07))'
```

MattPark:
Same result:
[~]
[root@nas]# /opt/bin/bash -c 'echo $(printf "%d" $(echo 99))'
99
[~]
[root@nas]# /opt/bin/bash -c 'echo $(printf "%d" $(echo 09))'
/opt/bin/bash: line 0: printf: 09: invalid number
0
[~]
[root@nas]# /opt/bin/bash -c 'echo $(printf "%d" $(echo 08))'
/opt/bin/bash: line 0: printf: 08: invalid number
0
[~]
[root@nas]# /opt/bin/bash -c 'echo $(printf "%d" $(echo 07))'
7

[...]

PR #89, opened 2015-08-05
Query fixes

jjchico:
Simplifies number of users querying. Also correct printf 0-padding format to better reflect the original intention.
May solve issue #65

### DIFF
--- a/mscs
+++ b/mscs
@@ -1622,7 +1622,7 @@ queryStatus() {
     TOKEN=$(querySendChallengePacket $1 | cut -f 3)
     if [ -n "$TOKEN" ]; then
       # Use the challenge token for the packet.
-      PACKET=$(printf "%.8x" $TOKEN)
+      PACKET=$(printf "%08x" $TOKEN)
       # Send the information request packet to the Minecraft query server.
       RESPONSE=$(querySendPacket "$1" "00" "$PACKET" "Cl>Z*Z*Z*Z*Z*s<Z*")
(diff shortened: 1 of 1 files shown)


jjchico:
Thanks for the merge.

sandain:
Thanks for the patches!

mscs
```diff
@@ -1965,7 +1979,7 @@ case "$1" in
     if serverRunning "$2"; then
       # If the world server has users logged in, announce that the world is
       # being deleted.
-      if [ $(printf "%d" $(queryStatus $2 | cut -f6)) -gt 0 ]; then
+      if [ $(queryNumUsers $WORLD) -gt 0 ]; then
```
sandain: Hmm, I didn't look very closely before accepting this patch.  Shouldn't this line be:
```
      if [ $(queryNumUsers $2) -gt 0 ]; then
```

mscs
```diff
@@ -1991,7 +2005,7 @@ case "$1" in
     if serverRunning "$2"; then
       # If the world server has users logged in, announce that the world is
       # being disabled.
-      if [ $(printf "%d" $(queryStatus $2 | cut -f6)) -gt 0 ]; then
+      if [ $(queryNumUsers $WORLD) -gt 0 ]; then
```
sandain: same here

sandain:
I just noticed your new pull request.  This issue is fixed there.

PR #90, opened 2015-08-05
World state checking and other minor fixes

jjchico:


jjchico:
checkPermissions is always executed since [0d7a3de905697c0a39482c27c157011230e0a573](https://github.com/sandain/MinecraftServerControlScript/commit/0d7a3de905697c0a39482c27c157011230e0a573). I introduced the change so that old-style disabled worlds could be migrated automatically. So, effectively, since that change all operations (including list/ls) are restricted to root/USER_NAME. Universal list/ls permissions can be activated again but the solution is not very elegant: force permissions everywhere except for list + migrate old worlds only if running the right user.

In fact I have a proposal for a different permission handling system that will make this and other related problems irrelevant, but we better comment that as an issue/proposal.

sandain:
I just checked, that commit broke the bash completion script.  The ls command needs to work without being root or the minecraft user.
Here is what happens when I try to type `sudo mscs start`, then use tab completion to see a list of available worlds to start:
```
$ sudo mscs start 
Expected  find      the       to        user      WARNING   
```

Expected would look like:
```
$ sudo mscs start 
alpha    beta    lobby
```

jjchico:
Yes, I understand the problem. I always "su - minecraft" so I did not notice. My proposal at issue #91 would solve it because you may try any operation either privileged or unprivileged.
In any case, "sudo mscs start" would not be de ideal solution since it still would execute most of the script as root. "sudo -u minecraft mscs start" would be better, but clumsy.

The proposal also handles this. E.g., you could do "sudo mscs start" and still get all the script executed as "minecraft" not just the operations delegated to "execute".

In fact, you may allow, for example, any user in the group "minecraft" to execute "sudo mscs <op>" as the user minecraft, not just administrators (users in group "sudo").

Please check the proposal. I am very interested in you opinion.

jjchico:
In the meantime I can resurrect the unprivileged list/ls if you wish.

sandain:
I would like to see the unprivileged `ls` resurrected while we discuss Issue #91.  Can you handle that, or do you want me to put it together?

I like the changes in patches e8068441c8fccb2c89adb25100bcd8aa7f9943a8 and 8d07b6414e13b5b2086f042cad17d6997d0dca92, and I will accept 61f097d090466a96c2f090c6e0caad46d3068d42 with the spelling fix.  Do you want to submit a new pull request with just those patches?

jjchico:
OK. I'll fix the spelling and do a new PR.

Issue #91, opened 2015-08-06
Better privilege handling


jjchico:
## Rationale

Currently, mscs can only be executed by root or a privileged user $USER_NAME (defaults to "minecraft"). Permission are assured by the checkPermissions() method in the script that forces the script to be executed only by root or $USER_NAME and by the execute() method, that executes all critical tasks as $USER_NAME even if the script is executed by root, implementing a kind of privilege dropping solution.
This approach works correctly in general but has a number of limitation in my modest opinion:

1. A dedicated user is always needed to run the script. A competent system administrator my set-up adequate folder and group permissions so that a group of user my control the servers, but this is explicitly blocked by the script.
1. When executed as root, even if critical tasks are handled by the "execute" method, the rest of the script is still executed by root with a risk of severe system damage.
1. Having to pipe all significant work through "execute" is annoying, development time consuming, low performing and ....... ugly? Still it does not eliminates a severe security risk (see above).

## Proposal

My proposal is to add a small wrapper script that sorts out privileges and executes the real control script with the unprivileged user. The new scripts would be:

* mscs: wrapper script named after the original full script for compatibility. It would behave exactly as the current "mscs". It will read $USER_NAME and executes the real script through "sudo" completely unprivileged.
* msctl (or mscs.real, or mscs.user, ...): The current script but made completely user-agnostic. In particular, "execute" and "checkPermissions" could be dropped.

## Benefits

* Solves the current limitations cited above.
* mscs.real could be used if no permission handling is needed or if it is done by the sysadmin. Also to execute unprivileged operations like "list", "status", etc. Ideal for administrators offering MC server hosting: user are free to run mscs.real as their user name. No further admin settings needed.
* More secure: the whole script would be executed unprivileged (no root).
* Better permission control: through "sudo", the sysadmin can control who can execute the script as the MC user, not just $USER_NAME or root.
* Completely transparent migration: "mscs" wrapper will act exactly as the current "mscs".
* Great code simplification in mscs.real by removing "execute".

## Drawbacks

* One more file in the project (5-10 lines of code).

I would be happy to implement it, but I would like to get some opinions from the community (more than one is a community, isn't it?)

sandain:
[...]

PR #94, opened 2015-08-12
Better privilege handling

jjchico:
See issue #91.

* The script itself is now named "msctl". It can be executed by any user without root privileges.
* Default location for server and world data is $HOME/mscs
* Configuration is read from $HOME/mscs.conf or $HOME/.config/mscs/mscs.conf
* "mscs" is now a wrapper that uses "msctl" and (should) retains compatibility with old "mscs": executes under "minecraft" user, uses "/opt/mscs" as default location, reads configuration from "/etc/default/mscs".
* There is no need to "sudo mscs" anymore: "mscs" can be executed by any user with enough privileges to run "msctl" as user "minecraft". For example, this lines can be added to the "sudoers" file to allow any user in group "minecraft" to execute "mscs" as user "minecraft" without asking for a password:

  Defaults:%minecraft	!authenticate
  %minecraft		ALL = (minecraft) /usr/local/bin/msctl

* Now, "start" will not create a default world if no world exists. It will tell the user how to do so using "create". It lead to some inconsistencies like creating a default world when all worlds were disabled.
* Command line parsing infrastructure have been created. Config file (-c) and data location (-l) can now be overridden through the command line (useful for "mscs" backwards compatibility).
* Other minor changes.

Hopefully, compatibility with current "mscs" has been retained.


jjchico:
With this PR a finish my particular "summer of code" with mscs. I have some additional changes in mind, including a new per-server configuration system, but they will break mscs compatibility and will need some manual migration. These additional changes  even including this PR) may better be started as an independent fork to avoid breaking current mscs installations. What do you think?

sandain:
Looks good, thank you for all this work.  I won't have a chance to test this until later this week, but I didn't seen any problems while looking through the code.  I didn't notice any changes to the README file; there are a few things that need to be updated before this goes live, but I can take a look at that when I do the final review.

jjchico:
I left the README until the new changes are tested.

sandain:
I left some comments in issue #91 for this pull request.

PR #112, opened 2016-01-31
Roflicide README patch -- updated

Roflicide:
* Removed incorrect information about editing files `mscs` and `msctl` files directly, and added info on creating configuration files.
* Merged a few sections into a "getting started guide" for easier reading.
* Wrapped lines to a max of 80 characters (as best as I could)
* Fixed wiki links to point to the master wiki--please note that clicking on some links, right now, will redirect to the main page as I'm unable to move the pages from my wiki to the master one (don't have permission to create pages).
* Fixed `PATH-TO-MSCS` to point to the default location; fixed some minor grammar and spelling issues.
* Added a logo to try to make the page more visually appealing and welcoming to first-time readers.
* Also, reverted the copyright year back to 2015 so it can be handled in a separate pull request.


README.md
```diff
@@ -1,101 +1,63 @@
-Minecraft Server Control Script
-=============================
-
-A powerful command-line control script for UNIX and Linux powered Minecraft servers.
-
-
-## Index
-* [Features](#features)
+<p align="center">
+  <img src="http://i.imgur.com/fUM7Ne8.png" />
```
zanix: I like the logo

README.md
```diff
@@ -111,38 +73,344 @@ requirements that this script has that may not already be in place:
...
+Navigate to the `MinecraftServerControlScript` directory that you just 
+downloaded. 
+Configuration can be done with the included Makefile in Debian and
+Ubuntu like environments by running:
+
+        sudo make install
+Then, type
+
+        chmod -R u+w /opt/mscs
+        chown -R minecraft:minecraft /opt/mscs
```
zanix: sandain: Do you think the make file should chown and chmod the files in `/opt/mscs` and this be moved to a troubleshooting section?

README.md
```diff
@@ -111,38 +73,344 @@ requirements that this script has that may not already be in place:
...
+  
+Where `world` is the name of the world you specify, 
+and `port` is the server port (by default, use 25565).
+`ip` is optional and will be used if you wish to create multiple worlds 
+across different servers. 
+
+Afterwards, simply start the server via `mscs start [world]` where `world` 
+is the name of the world. 
+
+**Finally, accept the EULA**.
```
zanix: >Finally, accept the EULA

I think this can just be "Accept the EULA"
Also if I remember correctly, you have to manually start the world again after editing `eula.txt`

README.md
```diff
@@ -111,38 +73,344 @@ requirements that this script has that may not already be in place:
...
+
+`eval echo ~$minecraft`
+
+Then simply `cd` into the directory it outputs and create the file.
+
+Once you've created the file, you need to populate it with a list of properties. 
+The list of properties can be found 
+[here](https://github.com/sandain/MinecraftServerControlScript/wiki/Global-Server-Settings).
+
+## Automated backups and restarts
```
zanix: I think this section would be better named "Scheduling Backups Restarts and Other Tasks"
And create another section with examples like mapping or the other available commands.

README.md
```diff
@@ -111,38 +73,344 @@ requirements that this script has that may not already be in place:
...
+
+### Scheduling backups
+MSCS uses [**cron**](https://en.wikipedia.org/wiki/Cron),
```
zanix: Better worded as "To schedule tasks such as backups, you need to use cron, a..." or something like that.

README.md
```diff
@@ -111,38 +73,344 @@ requirements that this script has that may not already be in place:
...
+
+### Scheduling restarts
+You can schedule restarts for the server following the same method as outlined 
+in [scheduling backups](#scheduling-backups).  
+Simply change the scheduled command from `backup <myWorld>` to 
+`restart <myWorld>`. `myWorld` 
+is the name of world you wish to restart; omit if you wish to 
+restart all worlds.
+
+## Mapping the world
```
zanix: This can be added in the future, but this section should mention `overviewer-settings.py`
By default, mscs creates this file with defaults but it can be further customized for each server.

zanix:
Also, it seems that Roflicide cannot create pages.
Roflicide , while we wait for sandain to figure permissions out, can you give me a list of page names to create?

README.md
Roflicide: Thank you!:)

README.md
Roflicide: Yea, I should have made that more clear.

README.md
Roflicide: Sounds good, though when you say create another section with mapping and other available commands, what do you mean? B/c I did make a mapping section & there is a command list section.

README.md
Roflicide: Okay:)

README.md
Roflicide: Yes, good idea.

README.md
zanix: Just mentioning that you can schedule other tasks using cron such as mapping.

Roflicide:
here's a few I think we need so far:
* List of Global Server Properties
* Manual Configuration
* Enabling Forge, BungeeCord, SpigotMC, and other minecraft versions and software

README.md
Roflicide: Ohhh--yes. I guess I was under the assumption the map auto-updated (I used to use DynMap back in the day, that's probably why). Yea, that should be added too.

zanix:
Created:
https://github.com/sandain/MinecraftServerControlScript/wiki/Global-Server-Settings
https://github.com/sandain/MinecraftServerControlScript/wiki/Manual-Configuration

I noticed the forge, bungeecord, and spigotmc already linked to an existing page.
https://github.com/sandain/MinecraftServerControlScript/wiki/Server-Customization-Examples

Roflicide:
Okay great, thanks for making those pages!
Two questions for you:

1) Why is there two locations of MSCS? Like theres one where I downloaded
it and theres one located in /usr/local/bin, and it looks like they're the
same in both locations.
2) Let's say I want to run a cron job every second on the map so that it
dynamically (or almost dynamically) updates. If I start a cron job, and
then a second later another job commences before the first one is completed
(i.e., that job would be rendering the map) what would happen--would it
wait for the job to complete then run, or would it crash?

I guess what would be a better question is how can I make cron do jobs
automatically, once the last job finishes, considering that the time
intervals for each job won't be constant? Is this even possible to do with
cron?

I'm just trying to make the map update in a semi-dynamic fashion.

zanix:
1) The one in `/usr/local/bin` is created when you run `sudo make install`
After you run `sudo make install`, you can actually remove the downloaded files.

The make file is pretty easy to follow, but I can summarize what it does:
https://github.com/sandain/MinecraftServerControlScript/blob/master/Makefile
* Copy `msctl` and `mscs` to `/usr/local/bin`
* Copy the mscs completion file to `/etc/bash_completion.d/mscs`
* Create the minecraft user
* Link `/usr/loca/bin/mscs` to the system startup (/etc/init.d for upstart or `mscs.service` for systemd)

2) Cron doesn't allow you to run every second, the lowest is every minute.
Also, Overviewer takes a while to run, and calling it multiple times runs a new instance so yes you system might crash.
Handling this case is probably better asked over in Overviewer's issue queue or IRC channel.

If you want a dynamic map, I would suggest just using Dynmap.

README.md
sandain: I do too.  Nice!

README.md
sandain: I thought the Makefile would create the correct permissions, but I could be wrong.  I like the idea of adding this to a troubleshooting section

sandain:
Roflicide, you should have access to modify wiki pages.  Feel free to do whatever you want.  I'll try to look over changes.  We will get these README changes commited soon, I just need to look over them again.

README.md
zanix: I would have assumed that as well.
When the makefile runs `adduser`, that should set `/opt/mscs` as owned by `minecraft`. When mscs is ran, it should also create all the directories as the `minecraft` user, which should set the owner as `minecraft`.

README.md
```diff
@@ -111,38 +73,344 @@ requirements that this script has that may not already be in place:
 If you are running Debian or Ubuntu, you can make sure that these are
 installed by running:
 
-    sudo apt-get install default-jre perl python wget rdiff-backup socat iptables
+`sudo apt-get install default-jre perl python wget rdiff-backup socat iptables`
```
sandain: Everywhere else we talk about running a command, it has spaces before the command rather than using the back-tick marks around the command.  Since the back-ticks and spaces have the same affect (placing the code in a box), I would prefer to keep this with spaces to be more readable from the command line.  If there weren't enough spaces to get this in a box (it looks like it may have been short), there should be other examples in the README that have the correct number of spaces.

README.md
```diff
@@ -111,38 +73,344 @@ requirements that this script has that may not already be in place:
...
+just created. 
+So I would drag the world `world` into the `vanillaMC` folder. 
+The path of `world` (the actual world folder) would now be 
+`/opt/mscs/worlds/vanillaMC/world/`.
+
+The finished file structure should resemble the following, with a world named 
+`world` and a containing folder name "vanillaMC":
+````
+/opt/mscs/vanillaMC       // The path
+
```
sandain: this should read /opt/mscs/worlds/vanillaMC

README.md
```diff
@@ -111,38 +73,344 @@ requirements that this script has that may not already be in place:
...
+from the last example). 
+Then, accept the EULA, run the start command again, and you're good to go!
+
+As a last note, make sure you check `server-port` and `query.port` in 
+`server.properties` 
+to make sure it does not overlap with other servers created by the MSCS script. 
+Also ensure that `enable-query` is set to `true`.  
+If you do not have `enable-query` and a `query.port` set, 
+you will not be able to check the status of the world with the script.
+
```
sandain: I don't think that this sequence that you have laid out would actually work.  Did you test it?

If I read the directions correctly, the `server.properties` file is in `/opt/mscs/worlds/vanillaMC/`, and the world data is in `/opt/mscs/worlds/vanillaMC/world`.  However, when you run `mscs create`, it changes the `level-name` value in `server.properties` to `vanillaMC` from `world`, so that when Minecraft starts running it won't find the world data since it expects to find it in `/opt/mscs/worlds/vanillaMC/vanillaMC`.

Making a copy of a world and renaming the world are two different problems.  Renaming a world requires that both the `server.properties` file change and the location of the world data matches that change.  Perhaps I should add a rename command to the script to simplify this.

README.md
zanix: Look at the Mastering Markdown page under Examples and the Code tab
https://guides.github.com/features/mastering-markdown/

README.md
Roflicide: Now that I come to think of it, I never tested it directly and you're right. I had mistakenly dragged the world data folder into the `/worlds` folder without creating a containing folder, so I guess I assumed the containing folder could be any name (which is wrong, my bad). 

I edited the documentation to make the names match

README.md
Roflicide: Yea, I updated back to it's original form, I just had tried to condense the line down to 80 characters (that's why I added the tick marks since it's shorter than 4 spaces) but I realize it's a bit dumb now that I think about it when the rest of the code formatting uses 4 spaces.

README.md
```diff
@@ -282,118 +551,10 @@ ie.
...
-    server.properties
-    white-list.txt
-
-You can just copy your world into the worlds directory:
-
-    mkdir /opt/mscs/worlds/alpha
-    cp -R * /opt/mscs/worlds/alpha
-
-Make sure you check `server-port` and `query.port` in `server.properties` to make sure it does not overlap with other servers created by the MSCS script. Also ensure that `enable-query` is set to `true`.  If you do not have `enable-query` and a `query.port` set, you will not be able to check the status of the world with the script.
-
```
sandain: The directions I wrote here are correct.  We can expand upon them, and talk about how to rename a world the hard way (since I haven't written any code for an easy way yet).

This is untested, but something like this would work to rename `alpha` to `vanillaMC`:
    mkdir /opt/mscs/worlds/vanillaMC
    cp -R * /opt/mscs/worlds/vanillaMC
    mv /opt/mscs/worlds/vanillaMC/alpha /opt/mscs/worlds/vanillaMC/vanillaMC
    mscs create vanillaMC 25565

Roflicide: Okay, I copied yours back in, and included the part on renaming the folder.

README.md
Roflicide: Hmmm... could it be possible it would function differently on a different computer? When I ran `sudo make install` I had to `chmod` it in order for it to create the folders.

I'll create and add it to a troubleshooting section, though

README.md
```diff
@@ -1,148 +1,431 @@
...
-                 preinstalled.
-* Python       - Required by the Minecraft Overviewer mapping software.
-* GNU Wget     - Allows the script to download software updates via the
-                 internet.
-* rdiff-backup - Allows the script to efficiently run backups.
-* Socat        - Allows the script to communicate with the Minecraft server.
-* Iptables     - Although not explicitly required, a good firewall should be
-                 installed.
+### Importing an existing world
+### Import Existing Worlds
```
sandain: You have duplicate titles here.  The Index above links to `importing-an-existing-world`, but either wording is fine with me.

README.md
```diff
@@ -1,148 +1,431 @@
...
+Please visit the [wiki]
+(https://github.com/sandain/MinecraftServerControlScript/wiki/Server-Customization-Examples) 
+for additional information.
+
+### Adjusting global server settings
+In tandem with the `mscs.properties` file--which handles options for individual 
+worlds--
+is the `mscs.conf` or simply `mscs` file, which handles global server settings.
+
+**You must create this file**. By default, MSCS looks for this file in three
```
zanix: **You must create this file**. By default, MSCS looks for this file in three 

Might want to specify that this file only needs to exist if you want to alter the default settings.

README.md
```diff
@@ -1,148 +1,431 @@
...
-You can [download](http://overviewer.org/downloads) premade binaries for
-supported systems, or build your own binary from source if needed.
+You can just copy your world into the worlds directory:
 
-Repositories for automatic installation are also available:
-* [Debian/Ubuntu](http://overviewer.org/debian/info)
-* [RHEL/CentOS/Fedora](http://overviewer.org/rpms/info)
+    mkdir /opt/mscs/worlds/alpha  // Makes a directory called `alpha` in the `/worlds/` folder
+    cp -R * /opt/mscs/worlds/alpha // Copies the `alpha` world data folder into `/opt/mscs/worlds/alpha`
+
```
sandain: I don't like having the comments in line on commands that people are going to want to copy-paste.  As is, if somone were to copy-paste your mkdir command, they would end up with directories called `Makes`, `a`, `directory`, etc in whatever directory they happen to be in.

 Also, I think it would be best to mention running `mscs create alpha` to setup the `server.properties` and `mscs.properties` files at this point

README.md
```diff
@@ -1,148 +1,431 @@
...
+Afterwards, start the server via `mscs start [world]`. 
+Then, accept the EULA, run the start command again, and you're good to go!
+
+As a last note, make sure you check `server-port` and `query.port` in 
+`server.properties` 
+to make sure it does not overlap with other servers created by the MSCS script. 
+Also ensure that `enable-query` is set to `true`.  
+If you do not have `enable-query` and a `query.port` set, 
+you will not be able to check the status of the world with the script.
```
sandain: A lot of repetition here.  If the create, copy and rename all mention running `mscs create`, there is no need to include another section.  The EULA section should probably be right there with create, since copy and rename are probably functional servers already.  Since we are telling users that they should run `mscs create`, then we probably don't have to also mention verifying the `server-port`, `query.port`, `etc` since that is the job of `mscs create`.

zanix: I think the part about verifying the ports comes from importing a world to be controlled by the script. We could say that users should run `mscs create <world>` first, then merge in the settings from the existing world into `server.properties`.

README.md
```diff
@@ -1,148 +1,431 @@
...
+* $CLIENT_VERSION - The version of the client software.
+* $SERVER_VERSION - The version of the server software.
+* $SERVER_JAR - The .jar file to run for the server.
+* $SERVER_ARGS - The arguments to the server.
+* $INITIAL_MEMORY - The initial amount of memory for the server.
+* $MAXIMUM_MEMORY - The maximum amount of memory for the server.
+* $SERVER_LOCATION - The location of the server .jar file.
+
+#### Example key/value pairs
+
```
sandain: I think it is important to note that these example key/value pairs are also the default values and will be used if missing from `mscs.properties`

Roflicide: Okay, I incorporated in the recommendations

README.md
```diff
@@ -1,148 +1,415 @@
...
+(https://github.com/sandain/MinecraftServerControlScript/releases).
 
-* Get the development version as a [zip file](https://github.com/sandain/MinecraftServerControlScript/archive/master.zip):
+* Get the development version as a [zip file]
+(https://github.com/sandain/MinecraftServerControlScript/archive/master.zip):
 
-        wget https://github.com/sandain/MinecraftServerControlScript/archive/master.zip
 
-* Make a clone of the [git repository](https://github.com/sandain/MinecraftServerControlScript.git):
+     wget https://github.com/sandain/MinecraftServerControlScript/archive/master.zip
```
zanix: Indenting this doesn't seem to put it in a code block, could be the markdown parser thinks it's a new line for the list item above. Might have to wrap this in ```

README.md
```diff
@@ -1,148 +1,415 @@
...
 
     sudo make install
 
-You can manually add the `minecraft` user and install the script with the
-following commands:
+   
+This will give the user you created in the config 
+(by default, the user `minecraft`) 
+access to write in the `/opt/mscs` folder. 
+If you configured MSCS manually when you installed the script,
```
zanix: > If you configured MSCS manually when you installed the script...

This line should probably go into the install/configure manually section.

zanix:
This is coming along nicely, good work @Roflicide

README.md
Roflicide: Ok

README.md
Roflicide: Good catch, seems i had to add another indent for it to recognize it

Roflicide:
@zanix Thank you kind sir!

Roflicide:
Also, I know you mentioned that the downloaded directory is unecessary after you run `sudo make install`--perhaps I should add in a bit on deleting the downloaded directory in the Installation section? Otherwise, I think it will cause confusion for some people

zanix:
I leave mine alone and run `git pull` then `sudo make update` when I want to update the script.
Think of the git download as an installer file, like an exe on Windows.
We could tell have a how to update section with these steps, and if they deleted the download they can run `git clone` again.

Roflicide:
Ahhh okay that makes more sense now, I'll add in an update section

zanix:
You might want to add that you run these commands in the folder you downloaded the script from git.

Roflicide:
Ok, added that bit in. 

If you don't mind, I have a quick off topic question: I tried doing 
`find / -type d -name MinecraftServerControlScript | cd` 
which didn't work; this led me to the difference between programs with `STDIN` and programs that just have command line arguments, like `cd`, which is why the above command didn't work since `cd` doesn't have an input stream.

My question is, do you know why `cd` doesn't have an input stream like any of the other commands? The best [answer I could find](http://unix.stackexchange.com/questions/50022/why-cant-i-redirect-a-path-name-output-from-one-command-to-cd?rq=1) said that 

> cd is not an external command - it is a shell builtin function. It runs in the context of the current shell, and not, as external commands do, in a fork/exec'd context as a separate process.

However, to me this doesn't really explain it at all: shouldn't all commands, even shell commands, have a `STDIN` stream because they need to read in data? Why is `cd` and a select few others special?

zanix:
I would say just because it doesn't
However you could possibly do this `cd $(find / -type d -name MinecraftServerControlScript)` although I'm not sure what would happen if it found multiple locations

sandain:
This is looking good.  Thanks @Roflicide!  If you are ready for a final review, I'll take one more closer look.  If I don't see anything major, I'll merge these changes.  After that, I can make any minor edits that I think it needs.

I don't have an explanation to you RE the `STDIN` business.  You can capture output from a command for a similar affect:
```
MSCS=$(find / -type d -name MinecraftServerControlScript)
cd $MSCS
```

I'm not sure what would happen if `find` happens to find more than one directory, so YMMV.

Roflicide:
@sandain Thank you! Yea, whenever you can look over it please go ahead--if you need me to change anything, let me know and I'll be more then happy to do so. :)

@sandain @zanix Thanks for the responses! In terms of the `STDIN` question, I think you're right when saying it's just apart of the design philosophy of Linux/Unix; the best answer I could find was [Why do some commands not read from their standard input?](http://unix.stackexchange.com/questions/140522/why-do-some-commands-not-read-from-their-standard-input), but even that was very murky on the details surrounding why some commands get `STDIN` and others do not.

In terms of the find command, I tested it out by making two directories:

`/usr/myDir` & `/home/myDir` 
and it always kept navigating to the `/usr/myDir` directory for some reason--not sure why. I suspect it has to do something with where it was programmed in the program to look first.

sandain:
I just got done reviewing all of the changes.  It looks great!  Thanks for doing this @Roflicide.

zanix:
Yes, awesome effort. Thanks @Roflicide!

Roflicide:
@sandain @zanix Thanks guys!

sandain:
It was bugging me that we had text about BungeeCord install notes on the wiki that weren't actually there.  So I went ahead and wrote some [install notes](https://github.com/sandain/MinecraftServerControlScript/wiki/Server-Customization-Examples#bungeecord) to add to the wiki.  If anyone gives this a try and the instructions don't work for you, let me know.

Issue #102, opened 2015-11-04
How to use the msctl script


sandain:
There is currently no documentation on how to use the `msctl` script, just the `mscs` script.

Roflicide:
Is this covered by the recent updates to the README or no?

sandain:
No.  The `msctl` script can be configured to work for multiple users on the same system, with each user able to control their own servers but not servers controlled by others.  So, for example lets say I had a user account on my server named `sandain`, if I were to use the `msctl` script the world files would be stored in `/home/sandain/mscs/worlds` instead of the global location `/opt/mscs/worlds`.  User `bob` on my server could independently run servers out of his home directory `/home/bob/mscs/worlds`, and neither would conflict with servers run by the `minecraft` user in `/opt/mscs/worlds`.  All users still have to coordinate which ports are being used, but @jjchico solved the permissions problems to make this possible.

The `mscs` script is now simply a wrapper for the `msctl` script that sets up the `minecraft` user to be used, so most of the documentation is exactly the same for the two scripts.

PR #115, opened 2016-02-17
Update README.md

Roflicide:
This patch is addressed to fix [Issue #102](https://github.com/sandain/MinecraftServerControlScript/issues/102).

* Added a section on using the MSCTL script (called it "configuring multiple users") and created a wiki page. I think I got the idea down, though there might be some things you want to change/add.
* Edited the "adjusting global server settings" section in the README. To make it simpler, I simply wrote that the user should create the file in the `/etc/default/` folder and call it `mscs`, since the other two locations are more important for user-by-user (I moved these with instructions to the "configuring multiple users" wiki page).

@sandain I saw you edited some parts on the wiki, it's looking good! I haven't got to reading the page  you added some BungeeCord install notes to, but I will definitely do in the near future and give you any feedback! (as I'm adding BungeeCord to my server soon!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
     * [Default world properties](#default-world-properties)
     * [Enabling Forge, BungeeCord, and other server software (optional)](#enabling-forge-bungeecord-and-other-server-software-optional)
   * [Adjusting global server options (optional)](#adjusting-global-server-settings-optional)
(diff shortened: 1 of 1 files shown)


zanix:
Looks pretty good to me

sandain:
Looks good. Thanks

PR #119, opened 2016-02-21
Run overview from backup directory

egut:
A small patch that run overview in backup directory instead of world directory so we don't have to run the server in save-off that long. The reason is that sometime the map generator sometimes (especially at reruns) can take hours and having save-off for hours could be very bad ;)


sandain:
This seems like a good idea.  Thanks for the patch.

PR #120, opened 2016-02-21
Dockerify MSCS

egut:
As I'm a fan of docker and think mscs is really good so I placed mscs and all its dependencies in a docker container. 

This also need some small tweaks to mscs to enable it run in foreground and not background as you would normally do. So now there are a one extra option -f  that starts the minecraft server in the foreground (default background).

I also added 2 dockerfiles show casing how to set it up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,14 @@ MSCS_INIT_D := /etc/init.d/mscs
 MSCS_SERVICE := /etc/systemd/system/mscs.service
 MSCS_COMPLETION := /etc/bash_completion.d/mscs
 
-.PHONY: install update clean
+.PHONY: docker-build docker-run install update clean
+
+docker-build:
+	docker build -t egut/mscs .
+	docker build -t myworld myWorld
+
+docker-run: docker-build
+	docker run -t -p 25565:25565 --name=myworld -d myworld
 
 install: update
 	adduser --system --group --home $(MSCS_HOME) --quiet $(MSCS_USER)
--- a/mscs
+++ b/mscs
@@ -22,18 +22,44 @@ if [ -r $MSCS_DEFAULTS ]; then
   MSCS_ARGS="-c $MSCS_DEFAULTS "
 fi
 
+# gosu for easy step-down from root
+# see: https://github.com/tianon/gosu
+GOSU=${GOSU:-"/usr/local/bin/gosu"}
+
 # Setup the default user name.
 USER_NAME=${USER_NAME:-minecraft}
 
 # Setup the default installation location.
 LOCATION=${LOCATION:-"/opt/mscs"}
 
+#Check permissions LOCATION and some top directories
+if [ -n "$(find $(find $LOCATION -maxdepth 2) \! -user "$USER_NAME" -print -prune -o -prune)" ]; then
+  echo "Note: Wrong owner of the directory, fixing"
+  if [ "$(whoami)" = "root" ]; then
+    chmod -R u+w /opt/mscs && chown -R minecraft:minecraft /opt/mscs
+  else
+    sudo chmod -R u+w /opt/mscs && chown -R minecraft:minecraft /opt/mscs
+  fi
+fi
+
 # Setup the arguments to the msctl script.
 MSCS_ARGS="-p $PROG -l $LOCATION $MSCS_ARGS $@"
 
 # Run the msctl script.
 if [ "$USER_NAME" = "$(whoami)" ]; then
+  # Correct user
   msctl $MSCS_ARGS
+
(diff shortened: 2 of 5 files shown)


msctl
```diff
@@ -1209,9 +1220,15 @@ start() {
...
+  if [ $RUN_IN_FOREGROUND -eq 1 ]; then
+    sh -c "tail -f --pid=\$$ '$WORLD_DIR/console.in' | {
+      $SERVER_COMMAND mscs-world='$1' > '$WORLD_DIR/console.out' 2>&1; kill \$$;
+      }"
+  else
+    sh -c "tail -f --pid=\$$ '$WORLD_DIR/console.in' | {
+      $SERVER_COMMAND mscs-world='$1' > '$WORLD_DIR/console.out' 2>&1; kill \$$;
+      }" &
+  fi
   # Verify the server is running.
```
sandain: How does this function with multiple worlds?  It seems like this would get stuck starting the first world in the list and wouldn't try starting any others until that first one exited.

egut: Great catch, I didn't think of that. Do you like to have a new pull-request with more or less only the dockerfiles? If you like I can also a docker-readme file with more examples.

Just for clarification when running inside docker container you would normally only run one world per container. So running multiple worlds inside one container wouldn't make that much seans. To run multiple worlds you would do something like:

    docker build -t mscs .
 
    <create the worlds volumes, init and setup the config for the worlds>

    docker run -d -t -p 25565:25565 -v world1:/opt/mscs --name=world1 mscs
    docker run -d -t -p 25566:25565 -v world2:/opt/mscs --name=world2 mscs

Makefile
```diff
@@ -7,7 +7,14 @@ MSCS_INIT_D := /etc/init.d/mscs
 MSCS_SERVICE := /etc/systemd/system/mscs.service
 MSCS_COMPLETION := /etc/bash_completion.d/mscs
 
-.PHONY: install update clean
+.PHONY: docker-build docker-run install update clean
+
+docker-build:
```
sandain: Since you placed docker-build and docker-run first in the list, it is now the default rather than install.  I'm not sure everyone wants to use docker.

sandain:
Sorry I have been so slow to respond here.  I wonder if the docker stuff would work better as a wiki entry rather modifying the Makefile and adding a couple of short text files.  Do you need the Makefile changes for this to work?

egut:
The changes in the makefile was only for convenience and colud defentlly be later in the make file if it interfere with other stuff. (I normally never just run make :D) but as you say it might be better for the wiki. 

How do you prefer to get the text?

mscs
```diff
@@ -22,18 +22,44 @@ if [ -r $MSCS_DEFAULTS ]; then
...
+#Check permissions LOCATION and some top directories
+if [ -n "$(find $(find $LOCATION -maxdepth 2) \! -user "$USER_NAME" -print -prune -o -prune)" ]; then
+  echo "Note: Wrong owner of the directory, fixing"
+  if [ "$(whoami)" = "root" ]; then
+    chmod -R u+w $LOCATION && chown -R minecraft:minecraft $LOCATION
+  else
+    sudo chmod -R u+w $LOCATION && sudo chown -R minecraft:minecraft $LOCATION
+  fi
+fi
+
```
sandain: This was added to the troubleshooting section of the readme, and the Makefile was updated to fix this problem on install.  I don't think this needs to be here.

mscs
```diff
@@ -22,18 +22,44 @@ if [ -r $MSCS_DEFAULTS ]; then
...
+elif [ "$(whoami)" = "root" ]; then
+  # Running as root, step-down to user
+  if [ -x $GOSU ]; then
+    $GOSU $USER_NAME msctl $MSCS_ARGS
+
+  else
+    su $USER_NAME -p -c "msctl $MSCS_ARGS"
+  fi
+
```
sandain: Root should have access to sudo, I think.  Is this really needed?

msctl
```diff
@@ -1485,7 +1490,6 @@ overviewer() {
   fi
   # Announce the location to access the world map to players.
   if [ $SERVER_RUNNING -eq 1 ]; then
-    sendCommand $1 "save-on"
     if [ $(compareMinecraftVersions $(getServerVersion $1) 1.7) -ge 0 ]; then
         sendCommand $WORLD 'tellraw @a {
           "text" : "",
```
sandain: I think I already accepted this hunk

sandain:
It looks like everything I'm willing to accept has already been accepted, everything else can go into a wiki page.  I can copy/paste what you have here, or I can give you write access to the wiki if you want to create the page.  How would you like to proceed?

Again, sorry for my slow response here.

PR #123, opened 2016-03-15
Update README.md

Roflicide:
Added information on `flock` command for cronjobs to prevent job overlapping

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,13 +375,9 @@ for more information.
 
 The backups will be saved in `/opt/mscs/backups`.
 
-If you're running multiple cron jobs (i.e. mapping the world, creating backups, etc) that may overlap eachother, it's a good idea to get a package called `flock` that will prevent the jobs from overlapping so errors aren't thrown.
-
-To install:
-
-    sudo apt-get install flock
+If you're running multiple cron jobs (i.e. mapping the world, creating backups, etc) that may overlap eachother, it's a good idea to use a package called `flock` that will prevent the jobs from overlapping so errors aren't thrown. It should be installed on all linux systems by default.
     
-Following the example above, make your entry look like so:
+Continuing the example above, make your entry look like so:
 
      0 */2 * * *  flock -n /tmp/fcj.lockfile /usr/local/bin/mscs backup myWorld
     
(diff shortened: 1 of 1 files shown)


sandain:
Would it be better to add `flock` to the appropriate places in the script?

zanix:
It seems like all this does is create a randomly generated blank lock file in `/tmp`. As an alternative to an additional dependency, the script could make it's own lock files.

zanix:
Reading more about lock files it seems that some are not blank and contain the PID of the running process in them.

zanix:
From: http://bencane.com/2015/09/22/preventing-duplicate-cron-job-executions/
```shell
PIDFILE=/path/to/mscs.pid
if [ -f $PIDFILE ]
then
  PID=$(cat $PIDFILE)
  ps -p $PID > /dev/null 2>&1
  if [ $? -eq 0 ]
  then
    echo "Process already running"
    exit 1
  else
    ## Process not found assume not running
    echo $$ > $PIDFILE
    if [ $? -ne 0 ]
    then
      echo "Could not create PID file"
      exit 1
    fi
  fi
else
  echo $$ > $PIDFILE
  if [ $? -ne 0 ]
  then
    echo "Could not create PID file"
    exit 1
  fi
fi

sleep 25d
rm $PIDFILE
```

Adding this to the commands in mscs will probably need more than just the PID in the file or filename.

Roflicide:
Yea, actually flock should be installed by default. and yea i think it does do more than simply creating a lock file, we should just use it instead of creating our own. in terms of incoporating that into the script i think thats a great idea.

sandain:
What commands need `flock` added?  Your README example shows backup.  What about mapping?  Are there any others?

zanix:
I would say anything that takes longer than a minute or so, or can break if another processes runs the same command.
So I would say backup, restore-backup, map/overviewer, sync, logrotate, and maybe keep a lock open for each world when it's started.

sandain:
I'm going to reject this patch in favor of adding `flock` directly to the script rather than change the wording of the README.  I opened issue #138 to hold further discussion on this subject.

PR #131, opened 2016-05-25
Use useradd for non-Debian systems

carmenbbakker:
`adduser` does not work for a large host of distributions. This should work universally.


zanix:
Looks good to me

sandain:
Looks good.  Thanks for the patch.

PR #132, opened 2016-05-25
Systemd service

carmenbbakker:
This pull request is twofold.

First, make sure that the service is run as the correct user and group rather than as (I presume) root.

Second, make sure `/usr/local/bin` is in the service's PATH variable. I don't know if this is an issue on Debian, but openSUSE does not have `/usr/local/bin` in its PATH for systemd services. This is a fairly easy fix for that.

### DIFF
--- a/mscs.service
+++ b/mscs.service
@@ -7,6 +7,7 @@ After=network.target
 [Service]
 User=minecraft
 Group=minecraft
+Environment="PATH=/usr/local/bin:/usr/bin:/bin"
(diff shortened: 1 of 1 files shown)


zanix:
I think the service runs as root, but when it calls runs the mscs commands it will su as the minecraft user. That said, it's probably a good idea for the service to start off as the minecraft user anyway.

sandain:
Both of these look like reasonable changes.  Thanks.

Issue #129, opened 2016-04-29
Security vulnerability: privilege escalation


DanielBaulig:
It's possible to hijack the msctl scripts user privileges and run arbitrary commands as the user msctl is running as. E.g.

    echo 'echo I can run whatever I want as `whoami`' >malicious.mscs.conf
    mscs -c malicious.mscs.conf list 

I don't think this is intentional and should probably be addressed.
But worse, with e.g. a weakly configured sudoers rule (e.g. allowing `sudo msctl` to any user), this can be even used to run arbitrary commands as an arbitrary user:

    USER_NAME=root -c malicious.mscs.conf mscs list

I think the following two steps should fix this. That said, I'm by no means a shell security expert.
1) Do not allow injection of `$USER_NAME` from the environment. Always set a default value for the case `$MSCS_DEFAULTS` does not exist
2) Do not source the configuration file. Use `IFS="=" read -r` or another approach that does not execute the content of the config file.

zanix:
Good catch. I don't think we have had anyone check for the security of this script.

We are already parsing values in `ini` format when we read in `mscs.properties` for each world. Maybe we could go that route instead of sourcing the configuration file.

sandain:
This is the reason I was worried about sourcing an outside file to begin with.  I agree, the best solution going forward is probably to get rid of the defaults file and move to using the code that reads in the `mscs.properties` file.  Unfortunately I have neither the time nor the will to tackle such a change at this time.  For now I will recommend only using this script in controlled environments where users can be trusted to not do such hijinks.

zanix:
Of course if someone were to make a pull request with the changes, I'm sure it would be accepted or worked on to make it acceptable.

sandain:
I will take the time to review pull requests.  I just don't have the time to make the required change to the code and test it to verify that the changes work.

If someone submits what looks like a reasonable fix that utilizes the parsing functions already in place and tells me that it has been tested and functions, I'll probably just accept it outright.

sandain:
Note that a reasonable fix must include updates to the README file, particularly the `Adjusting global server settings (optional)` and `Configuring MSCS for multiple users (optional)` sections.

sandain:
[...]

PR #134, opened 2016-06-18
New format for defaults file v4

sandain:
This removes an avenue of attack mentioned in Issue #129.

### DIFF
--- a/msctl
+++ b/msctl
@@ -391,12 +391,20 @@ getValue() {
   if [ -e "$1" ]; then
     # Find the key/value combo.
     KEY=$($PERL -ne '
-      $_ =~ s/[\x22\x27\r]//;
-      if ($_ =~ /^('$2')=.*$/i) { print lc $1; }
+      # Remove single and double quotes plus CR and LF.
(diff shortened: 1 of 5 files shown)


msctl
```diff
@@ -391,12 +391,20 @@ getValue() {
   if [ -e "$1" ]; then
     # Find the key/value combo.
     KEY=$($PERL -ne '
-      $_ =~ s/[\x22\x27\r]//;
-      if ($_ =~ /^('$2')=.*$/i) { print lc $1; }
+      # Remove single and double quotes plus CR and LF.
+      $_ =~ s/[\x22\x27\r\n]//g;
+      # Remove comments.
+      $_ =~ s/\x23.*//;
```
sandain: I wonder if this would be better as:
```
$_ =~ s/^\s*\x23.*//;
```

Let's say a someone has modified one of the URLs with a `#`, which isn't uncommon.  The code will currently assume everything after the `#` is a comment and will strip it.  If we require comments to be at the beginning of the line (with or without spaces first), we can get around this problem.  Shell scripts might allow a comment on the line, but I don't know if it is worth catching every possible configuration.

Issue #136, opened 2016-07-24
bungee server reports muliple versions and mscs script doesn't handle it properly


gee-one:
I'm not sure when this change happened, but it appears as if bungee servers are reporting multiple versions for compatibility.  Mine reports "1.8.x, 1.9.x, 1.10.x".  The issue arises when querying the status- it seems as if printf doesn't like the output and fails to parse it correctly.  This is running on Debian Jessie 8.5.  The actual minecraft servers are on a different machine.

```
root@bungee:/usr/local/bin# mscs status
Minecraft Server Status:
  bungee: /usr/local/bin/msctl: 1574: printf: 1.9.x,: not completely converted
/usr/local/bin/msctl: 1574: printf: 1.10.x: not completely converted
running version 1.8.x, (1 of 1 users online).
running version 0 (20 of 0 users online).
    Process ID: 31188.
    Memory used: 186632 kB.
  unplugged: disabled.
```

If I remove the reference to VERSION in msctl the output is as intended, without the version information.

```
root@bungee:/usr/local/bin# mscs status
Minecraft Server Status:
  bungee: running version XXX (0 of 20 users online).
    Process ID: 31188.
    Memory used: 186632 kB.
  unplugged: disabled.
```

I can't speak for anyone else, but I would rather see the number of players online and the list of players, than get an error because of the version number.  The script seems to work as intended on vanilla minecraft and Spigot servers.  I think bungee started advertising multiple versions and uses priorities to allow different versions of clients to connect to different servers.

Lastly, thanks for a great script!

sandain:
I accepted your patch.  Thank you.

PR #137, opened 2016-07-24
update msctl for multiple bungeecord versions

gee-one:
Updated msctl to be more compatible with bungeecord and multiple versions.

Bungeecord server reports multiple version numbers for compatibility with different versions of clients.  The output doesn't not get parsed properly and printf throws an error.

See issue #136


sandain:
Thanks for the patch.  I don't have the means to test it right now, but it looks correct to me.  I will go ahead and assume that you have tested this and that it fixes the issue you noticed. Patch accepted.

PR #148, opened 2017-01-03
fixing bug WORLDS_LOCATION var in start function

seldonPlan:
in `start()` a directory check starting with `WORLDS_LOCATION` was not properly referenced. added `$WORLDS_LOCATION` to fix


sandain:
Thanks for the patch.

PR #149, opened 2017-01-03
Feature Request - prevent backups from copying symlink files

seldonPlan:
When backing up a mirrored world, the current behavior is to simply backup everything in the world directory, including the symlinks to the mirrored directory. This has a few implications:

- the backup operation is likely not catching the most recent world changes (if any), since `worldBackup` is not actually doing a sync call before it does it's thing
- since the mirror directory is defined in the root `mscs.defaults` file (hence not part of the backed up files themselves), it could potentially be changed by the user causing an unintended side effect of their world not behaving properly from a restore backup operation
- a symlink can also cause some issues when restoring a backup to another machine as is not resolved properly

This request sidesteps these issues altogether by simply ignoring the symlink in the backup via these three changes:

1. the behavior of the backup operation to use `rdiff-backup`'s flag of `--exclude-symbolic-links`
2. `start` function now needs to account for the fact that the symlink world file (`$WORLDLOCATION/$1/$1`) may not be present. On a backup-restore operation only the (non-symlink) `$WORLDLOCATION/$1/$1-original` directory would be available
3. the `worldBackup` function needs to perform a sync operation immediately prior to calling `rdiff-backup`

### DIFF
--- a/msctl
+++ b/msctl
@@ -1098,7 +1098,7 @@ stop() {
   sendCommand $1 "end"
   # Synchronize the mirror image of the world prior to closing, if
   # required.
-  if [ $ENABLE_MIRROR -eq 1 ] && [ -d $MIRROR_PATH ]; then
+  if [ $ENABLE_MIRROR -eq 1 ] && [ -L "$WORLDS_LOCATION/$1" ] && [ -d $WORLDS_LOCATION/$1/$1-original ]; then
     syncMirrorImage $1
     # Remove the symlink to the world-file mirror image.
     rm -r "$WORLDS_LOCATION/$1/$1"
@@ -1137,7 +1137,7 @@ worldBackup() {
   fi
   # Synchronize the mirror image of the world prior to closing, if
   # required.
-  if [ $ENABLE_MIRROR -eq 1 ] && [ -d $MIRROR_PATH ]; then
+  if [ $ENABLE_MIRROR -eq 1 ] && [ -L "$WORLDS_LOCATION/$1" ] && [ -d $WORLDS_LOCATION/$1/$1-original ]; then
     syncMirrorImage $1
   fi
   # Determine if we need to exclude the mirrored symlink or not
--- a/msctl
+++ b/msctl
@@ -1037,7 +1037,11 @@ start() {
     # <world>-original directory within <world> we didn't stop cleanly.
     if [ -d "WORLDS_LOCATION/$1/$1-original" ]; then
       # Remove the symlink to the world-file mirror image.
-      rm -r "$WORLDS_LOCATION/$1/$1"
+      # if recently restored from backup, this symlink may or may not exist
+      if [ -L "$WORLDS_LOCATION/$1/$1" ]; then
+        rm -r "$WORLDS_LOCATION/$1/$1"
+      fi
+
       # Move the world files back to their original path name.
       mv "$WORLDS_LOCATION/$1/$1-original" "$WORLDS_LOCATION/$1/$1"
     fi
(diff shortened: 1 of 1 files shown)


sandain:
Thank you for the patch.  I don't use this feature, so I'm glad to have someone else looking it.  Your pull request seems like a reasonable change, but I am going to let it sit here for a little while to let other people take a look at it.

zanix:
I do not use the mirror feature either so it's nice to get a pair of eyes on it. Nice catch on syncing the mirror before running the backup.

I never thought about moving the backups to another server before. That would be a good use case for an export function that could create a tar.gz of a specific restore point (without increments) or the entire backup (with increments).

I use a couple of symlinks in each my world folders, so I can use a global whitelist and banlist.
I'm not sure if I have an issue with this or not, since my use case is simple. I'm not sure if anyone else is doing something else with symlinks or not. I could see a couple modded servers sharing a common mods directory though.

seldonPlan:
I will admit my use case is probably a bit (overly) complex than mscs' ideal use case :) I am currently playing around with "ephemeral" aws server deploys with modded minecraft, where I backup the world state to some external storage (in this case I use aws s3). I then am free to kill the server when it wont be used for a few days, and redeploy on-demand, pulling the backup and restoring the world. So I am probably jumping across servers way more than average. 

I actually wrote a script that bundles the latest backup out of `rdiff-backup` into a `.tar.gz` archive. That is what I am using to bundle the world backups before sending them off to s3. So, for me, having that functionality in the mscs script would be useful.

Looking at your global whitelist/banlist example, i think you are right, and doing `--exclude-symbolic-links` is probably too heavy-handed to deal with what should be a single ignored file with `--exclude **/$WORLDNAME`. I can adjust the pr to do that instead

zanix:
Very nice, good idea when AWS server to save some cash.

About exporting `rdiff-backup` into an exportable `tar.gz` I think that would make a good addition to the script. If you have some extra time, make a PR or you could put your script on gist and someone else might adapt it.

Maybe we can add config to include extra `rdiff-backup` parameters so we exclusions could be added?

sandain:
That does sound like an interesting use case.  If you have the time, I'm sure others might be interested in hearing how you have things setup using AWS.  Perhaps you could add a section to the [wiki](https://github.com/MinecraftServerControl/mscs/wiki/Server-Customization-Examples).

seldonPlan:
I've made the change to only look at the symlink mirror directory instead of all symlinks. If `$WORLDS_LOCATION/$1/$1` exists as a symlink, then it is excluded from the backup, otherwise it is included.

I've also snuck in another change that I think cleans up the sync process a little bit (let me know if you'd rather this be its own pr). Both `worldBackup()` and `stop()` use the following idiom to determine if a sync should occur:
```
if [ $ENABLE_MIRROR -eq 1 ] && [ -d $MIRROR_PATH ]; then
    syncMirrorImage $1
fi
``` 

but `syncMirrorImage()` never references `$MIRROR_PATH` directly and itself only cares about the locations within the worlds directory
```
cp -Ru "$WORLDS_LOCATION/$1/$1/"* "$WORLDS_LOCATION/$1/$1-original"
```

so i'd like to change the idiom to check for what the sync operation actually uses:
```
if [ $ENABLE_MIRROR -eq 1 ] && [ -L "$WORLDS_LOCATION/$1" ] && [ -d $WORLDS_LOCATION/$1/$1-original ]; then
    syncMirrorImage $1  
fi
```

seldonPlan:
@sandain I would like to formalize what i've been doing in wiki form :). When i get my process stable, I would be happy to write something up about it in the wiki. I'll share the scripts I've been using as well in case they can be of any use.

sandain:
This is looking pretty good. I think changing to using `EXCLUDE_OPTION` was a good minimal change that will avoid the issue with symlinks that @zanix mentioned.  I'm fine with the change to the checks for `syncMirrorImage`, although I wonder if would be simpler to add the check to the subroutine, e.g.:
```
syncMirrorImage() {
  if [ ! $ENABLE_MIRROR -eq 1 ] || [ ! -L "$WORLDS_LOCATION/$1" ] || [ ! -d $WORLDS_LOCATION/$1/$1-original ]; then
    return
  fi
  # Sync the world server.
```
I'm fine with it either way, but lets let this sit for a bit longer for more comments before I accept the PR.

sandain:
Since there have been no further comments, I'm going to go ahead and accept this.

PR #151, opened 2017-01-08
allow useradd to create home directory from install target, not the $(MSCS_HOME) target

seldonPlan:
It is useful to be able to work as the `minecraft` user that is created from the `install` target in the Makefile. 

However, because the `install` target runs the `$(MSCS_HOME)` target beforehand (thus creating the home directory beforehand) the `useradd` command fails to pull over useful shell default files (like `.bashrc`, `.profile`, etc...) and anything else that may be living in `/etc/skel`. I don't know if this was  intended behavior or not since the `--create-home` option is supposed to setup those files for a newly added user.

This PR contains two changes: 

- remove the call to the `$(MSCS_HOME)` target when running `make install`. The `MSCS_HOME` directory will still be created by the `useradd` command, and on most systems, i think this would be enough by itself. The `useradd` command would then properly setup the home directory with the system default files

- add the `-K UMASK=0022` option to the `useradd` command. This just ensures that on a system with some differing defaults it still sets up the mscs home directory with the intended permissions the same way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ UPDATE_D := $(wildcard update.d/*)
 
 .PHONY: install update clean
 
-install: $(MSCS_HOME) update
+install: update
 	useradd --system --user-group --create-home --home $(MSCS_HOME) $(MSCS_USER)
 	chown -R $(MSCS_USER):$(MSCS_GROUP) $(MSCS_HOME)
 	if which systemctl; then \
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ UPDATE_D := $(wildcard update.d/*)
 .PHONY: install update clean
 
 install: update
-	useradd --system --user-group --create-home --home $(MSCS_HOME) $(MSCS_USER)
+	useradd --system --user-group --create-home -K UMASK=0022 --home $(MSCS_HOME) $(MSCS_USER)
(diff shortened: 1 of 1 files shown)


seldonPlan:
This is another PR influenced by my weird use-case. I didn't see any obvious reason why the `$(MSCS_HOME)` target was even needed in the Makefile, so let me know if i missed something. Thanks

sandain:
This change looks fine to me.  This does leave the `$(MSCS_HOME)` target without a use, so it should also be removed.

For a little context, this target has been in the `Makefile` longer than `useradd` has been there -- this probably should have been done when that change was made.

sandain:
I think that `useradd` should also be taking care of ownership of `$(MSCS_HOME)`, so we'll can't probably drop the call to `chown` as well.

zanix:
`useradd` is used in the script and `adduser` is used in the manual install instructions.
Should the wiki be updated to match the script?

sandain:
I went ahead and dropped the call to chown.

We should update the wiki to use the same directions as used in the `Makefile`.

sandain:
I added a patch to drop the `MSCS_HOME` target, updated the [wiki](https://github.com/MinecraftServerControl/mscs/wiki/Manual-Configuration), and added another patch to the `Makefile` to match the way the wiki did things.

PR #153, opened 2017-01-10
Merging from source

seldonPlan:
* msctl: use the PERL variable
* allow useradd to create home directory from install target, not the $(MSCS_HOME) target
* add umask permissions to useradd command
* Makefile: drop the call to chown
* Makefile: drop the MSCS_HOME target
* Makefile: use install rather than cp to match wiki docs


seldonPlan:
sorry... trying to update my own fork... please disregard

PR #157, opened 2017-02-18
Fix and improve the world mirrors

Wicloz:
The current mirror system does not stop properly and is generally unreliable.

The stop and start functions have been changed to handle the starting and stopping of mirrors much more reliably (and actually work in the first place). The use of ramdrives as the mirror location has been taken into account. The stop and force-stop function have been improve to prevent issues with the mirrors being synchronized as the world is shutting down and saving. The syncMirrorImage function now uses rsync instead of cp. Backups have been slightly changed to work with these changes.

All changes are fully backwards compatible.

### DIFF
--- a/msctl
+++ b/msctl
@@ -1098,10 +1098,12 @@ stop() {
   sendCommand $1 "end"
   # Synchronize the mirror image of the world prior to closing, if
   # required.
-  if [ $ENABLE_MIRROR -eq 1 ] && [ -L "$WORLDS_LOCATION/$1" ] && [ -d $WORLDS_LOCATION/$1/$1-original ]; then
+  if [ $ENABLE_MIRROR -eq 1 ] && [ -L "$WORLDS_LOCATION/$1/$1" ] && [ -d "$MIRROR_PATH/$1" ]; then
     syncMirrorImage $1
     # Remove the symlink to the world-file mirror image.
-    rm -r "$WORLDS_LOCATION/$1/$1"
+    rm -f "$WORLDS_LOCATION/$1/$1"
+    # Remove the world-file mirror image folder.
+    rm -Rf "$MIRROR_PATH/$1"
     # Move the world files back to their original path name.
     mv "$WORLDS_LOCATION/$1/$1-original" "$WORLDS_LOCATION/$1/$1"
   fi
--- a/msctl
+++ b/msctl
@@ -1019,9 +1019,9 @@ start() {
   mkdir -p "$WORLD_DIR"
   # If the original level exists but the actual level doesn't,
   # we probably just restored a backup.
-  if [ -d "$WORLDS_LOCATION/$1/$1-original" ] && [ ! -e "$WORLDS_LOCATION/$1/$1" ]; then
+  if [ -d "$WORLD_DIR/$1-original" ] && [ ! -e "$WORLD_DIR/$1" ]; then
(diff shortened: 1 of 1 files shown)


sandain:
Thanks for the pull request!  I'm going to need some time to review these changes -- I like that you kept each change separate, that will make it easier.

msctl
```diff
@@ -1028,29 +1038,36 @@ start() {
...
+    fi;
+    # If the symlink still exists, the server was stopped outside of mscs.
+    # SO DONT RESET THE MIRROR, just keep using it.
+    if [ ! -L "$WORLD_DIR/$1" ]; then
+      # Remove the mirror directory just in case.
+      rm -Rf "$MIRROR_PATH/$1"
+      # Copy the world files over to the mirror.
+      cp -a "$WORLD_DIR/$1" "$MIRROR_PATH/$1"
+      if [ $? -ne 0 ]; then
+        printf "Error copying world data, could not copy to %s.\n" $MIRROR_PATH/$1
```
zanix: Should some additional help text be added so users know what they can do to fix this?

zanix: Nevermind, this is a simple cp error.

msctl
```diff
@@ -1115,11 +1137,20 @@ stop() {
...
+  # Wait for the server to shut down.
+  WAIT=0
+  while serverRunning $1 && [ $WAIT -le 20 ]; do
+    WAIT=$(($WAIT+1))
+    sleep 1
+  done;
   # Kill the process id of the world server.
   kill -9 $(getJavaPID "$1") > /dev/null 2>&1
+  # Properly clean up.
+  stop $1
```
zanix: Very nice catch, forcestop doesn't do the same tasks that stop does.

msctl
```diff
@@ -1115,11 +1137,20 @@ stop() {
...
 forceStop() {
+  local WAIT
   # Try to stop the server cleanly first.
-  stop "$1"
-  sleep 5
+  sendCommand $1 "stop"
+  sendCommand $1 "end"
+  # Wait for the server to shut down.
+  WAIT=0
+  while serverRunning $1 && [ $WAIT -le 20 ]; do
```
zanix: I like the additional wait time here to see if the server will shut off on it's own.

sandain:
I just got done reviewing this. The logic looks correct. Have you fully tested all of these changes?

My only concern is that we are now adding `rsync` as a dependency.  We should add a check similar to those found [here](https://github.com/MinecraftServerControl/mscs/blob/master/msctl#L1668) and add documentation to the [README](https://github.com/MinecraftServerControl/mscs/blob/master/README.md#required-programs).

zanix:
I think `rdiff-backup` depends on `librsync1` but not the entire `rsync` application.
I believe `rsync` comes default in most distributions anyway.

Wicloz:
I tested all the changes and am currently running this for my servers, with the mirror folder in a tmpfs. My world is still fine after multiple reboots and crashes.

sandain:
I'm going to accept this PR, but I am disappointed that my request for an update to the README and a check for rsync at the beginning of the script was ignored.  I'll go ahead and make those changes myself.

Wicloz:
Sorry about that, I read between the lines too much and thought you intended to make the changes...

sandain:
No worries! Thank you for getting this code straightened out.

Issue #158, opened 2017-02-20
mscs status "bad ps syntax"


shimniok:
Received "bad ps syntax" warning when running mscs status. 

Error:
```
$ mscs status
Minecraft Server Status:
  alpha: not running.
  shimniok: running version 1.11.2 (0 of 5 users online).
    Process ID: 7673.
Warning: bad ps syntax, perhaps a bogus '-'? See http://procps.sf.net/faq.html
    Memory used: 349480 kB.
```

Tracing msctl yields:
```
+ ps -a -U minecraft -o+  pid,comm,args -ww
/usr/bin/perl -ne if ($_ =~ /^\s*(\d+)\s+java.+mscs-world=shimniok$/) { print $1; }
+ PID=7673
+ printf %d\n 7673
+ PID=7673
+ ps -h -p 7673 -o rss
Warning: bad ps syntax, perhaps a bogus '-'? See http://procps.sf.net/faq.html
+ printf     Memory used: %d kB.\n 409456
    Memory used: 409456 kB.
+ exit 0
```

ps version:
```
$ ps --version
procps version 3.2.8
```

Server info:
Linux Mint 13
uname: 
`Linux nogales 3.2.0-23-generic #36-Ubuntu SMP Tue Apr 10 20:39:51 UTC 2012 x86_64 x86_64 x86_64 GNU/Linux
`

Let me know if I can help troubleshoot/test/whatever.

shimniok:
Further investionation:

It appears -h is the offending flag. If the intent is to strip headers, use `--no-headers` (on this system, anyway). 

So, changing line (190) from:
  `ps -h -p $PID -o rss`

to:
  `ps --no-headers -p $PID -o rss`

results in:

```
$ mscs status
Minecraft Server Status:
  alpha: not running.
  shimniok: running version 1.11.2 (0 of 5 users online).
    Process ID: 7673.
    Memory used: 459604 kB.
```

sandain:
The `--no-headers` option is synonymous with `-h` on my system (Ubuntu 17.04):
```
$ ps --version
ps from procps-ng 3.3.12
```
I checked a few other systems, and the oldest version of `procps-ng` I could find, `3.3.9` has the same `--no-headers` options.  I would be fine with making this change, hopefully there isn't some variant of `ps` that only works with the `-h` option.  Do you want to submit a pull request?

shimniok:
[...]

PR #159, opened 2017-02-22
getJavaMemory(): changed ps flag --no-headers

shimniok:
#158 
Remedies incompatibility of -h flag with procps (ps) version 3.2.8 on older Linux Mint 13 (3.2.0-23-generic based on Ubuntu 12.04). 

The --no-headers flag appears to be widely supported for turning off header display of ps output. 

Newer Debian, Ubuntu and Mint uses procps version 3.3.x which seems to be based on procps-ng rather than the older procps.

### DIFF
--- a/msctl
+++ b/msctl
@@ -187,7 +187,7 @@ getJavaPID() {
 getJavaMemory() {
   local PID
   PID=$(getJavaPID "$1")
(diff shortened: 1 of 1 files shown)


sandain:
Thanks for the patch!

PR #160, opened 2017-03-17
Update config to include Google Maps API key needed for Overviewer

Roflicide:
Hey guys,

Long time no see! Hope you all are doing well. I had some spare time so I figured I might as well push some new changes to MSCS. 

Google now requires a (free) API key to access their Maps API. Because of this, the Overviewer team have been working on changing Overviewer over to [leaflet](http://leafletjs.com/). For now, though Overviewer is sticking with Google Maps. As such, Overviewer now requires an API key, too. So, I went ahead and added an overviewer-key property and some other lines of code to integrate MSCS and Overviewer smoothly. 

The main line of code uses a `sed` command. I've adopted this `sed` line of code from [here](https://github.com/overviewer/Minecraft-Overviewer/issues/1308#issuecomment-235898484). You can also read more about the issue at that link.

I've also added more documentation to the mapping section, and tried to clarify writing as I saw fit. 

I've been using this configuration with my server and its worked like a charm, though please check my code as my programming skills are mediocre lol.

Thanks!
-Mike


Roflicide:
So, after some more testing and reading other people's comments, it look's like the fix I used above can create issues because it doesn't target the actual source of overviewer overwriting the index.html file.

(Zeromusta created a script here that address the issue and seems to be working on my server.)(https://github.com/Zeromusta/docker-spigot-overviewer/blob/master/rootfs/usr/local/bin/ov_keyfix) I'm not sure if you would rather switch over to the new leaflet overviewer library (which is still in development, and unstable) or try figuring out a patch with the Google Maps API involving a fix like the one above.

README.md
```diff
@@ -475,38 +475,70 @@ Where `<world>` is the name of the world you wish to map (omit for all
...
+paste the following default values 
+(see [adjusting global server properties](#adjusting-global-server-properties)):
 
     mscs-overviewer-bin=/usr/bin/overviewer.py
     mscs-overviewer-url=http://overviewer.org
+    mscs-overviewer-key=PASTE-YOUR-KEY-HERE
     mscs-maps-location=/opt/mscs/maps
     mscs-maps-url=my.minecraftserver.com
+    
+* `mscs-overviewer-bin`: The overviewer file. If you did the default Overviewer installation,
```
zanix: Maybe reword to
> This is the location for the Overviewer binary. If you manually installed Overviewer to another location, you can enter the location here.

README.md
```diff
@@ -262,10 +257,13 @@ that you wish the world to use:
 
 ### Adjusting global server properties
 Default values in the script can be overridden by adding certain properties to one
-of the `mscs.defaults` files. The `mscs.defaults` files can be found found in one
+of the `mscs.defaults` files. The `mscs.defaults` files needs to be created in one
 of three places depending on how the script is being used. When using the `mscs`
-script, the `mscs.defaults` file can be found at `/opt/mscs/mscs.defaults`. When
-using the `msctl` script in [multi-user mode](https://github.com/MinecraftServerControl/mscs/wiki/Configuring-MSCS-for-multiple-users), the `mscs.defaults` file can be found at either `$HOME/mscs.defaults` or `$HOME/.config/mscs/mscs.defaults`.
+script, the `mscs.defaults` file __**will need to be created*__ at `/opt/mscs/mscs.defaults`. When
```
sandain: you have 2 asterisks before, one after.  this leaves 1 asterisk in the text when I don't think you meant to.

msctl
```diff
@@ -1375,11 +1375,7 @@ overviewer() {
           },
           {
             "text" : "Overviewer",
-            "color" : "aqua",
-            "clickEvent" : {
-              "action" : "open_url",
-              "value" : "'$OVERVIEWER_URL'"
-            }
+            "color" : "aqua"
           },
```
sandain: Why drop the link back to Overviewer here?

msctl
```diff
@@ -1407,7 +1407,7 @@ overviewer() {
             "color" : "aqua",
             "clickEvent" : {
               "action" : "open_url",
-              "value" : "'$MAPS_URL'/'$1'"
+              "value" : "'$MAPS_URL'"
```
sandain: Why drop the world here?  As far as I know, we don't have an `index.html` generated for this location, so directory browsing would have to be turned on for this to work.

sandain: But maybe we should generate an index.html for this location ...

sandain:
This looks like a good update. Thanks for the work.  I've left a few comments above.

RE leaflet: I'm open to whatever you guys think is best here.  I'm too busy to even give this any thought, so I'll defer to your wisdom.

README.md
Roflicide: Ah yes, good catch

msctl
Roflicide: My bad, I had dropped it on my own local server because I didn't really think it was necessary, but we should leave it

msctl
Roflicide: My bad again, I had changed it locally for my own server to point to somewhere else

Roflicide:
Thanks @sandain. I'm going to wait a little on deciding what to do and see if they're coming out with any updates anytime soon

README.md
```diff
@@ -508,8 +508,7 @@ paste the following default values
     mscs-maps-location=/opt/mscs/maps
     mscs-maps-url=my.minecraftserver.com
     
-* `mscs-overviewer-bin`: The overviewer file. If you did the default Overviewer installation, 
-   leave this untouched.
+* `mscs-overviewer-bin`: This is the location for the Overviewer binary.                                                                                                                       If you manually installed Overviewer to another location, you                can enter the location here.
```
sandain: Weird formatting here -- extra white space.  When viewed as a web page it's not an issue, but it looks odd when viewed as a text file.

sandain:
I like the addition of the `overviewer-key` parameter, it seems like a good stop-gap until they get leaflet integrated.  This part of your patch can always be reverted in the future when it is no longer needed.

Roflicide:
@sandain Okay, let me just add the part to ensure the Overviewer source files get overwritten

README.md
```diff
@@ -475,38 +475,69 @@ Where `<world>` is the name of the world you wish to map (omit for all
...
+paste the following default values 
+(see [adjusting global server properties](#adjusting-global-server-properties)):
 
     mscs-overviewer-bin=/usr/bin/overviewer.py
     mscs-overviewer-url=http://overviewer.org
+    mscs-overviewer-key=PASTE-YOUR-KEY-HERE
     mscs-maps-location=/opt/mscs/maps
     mscs-maps-url=my.minecraftserver.com
+    
+* `mscs-overviewer-bin`: This is the location for the Overviewer binary.                                                                                                                       If you manually installed Overviewer to another location, you                can enter the location here.
```
zanix: Odd white space here

README.md
```diff
@@ -475,38 +475,69 @@ Where `<world>` is the name of the world you wish to map (omit for all
 worlds).
 
 ## Mapping the world
-Minecraft Server Control Script uses [Overviewer]
-(http://docs.overviewer.org/en/latest/) to generate maps. After [installing]
-(#mapping-software), modify the settings (if necessary) found in the
-`mscs.defaults` file (see [adjusting global server properties]
-(#adjusting-global-server-properties)):
+Minecraft Server Control Script uses [Overviewer](http://docs.overviewer.org/en/latest/) to generate maps.
```
zanix: I like this documentation change. Nice info.

README.md
```diff
@@ -475,38 +475,69 @@ Where `<world>` is the name of the world you wish to map (omit for all
...
-
-### Adjusting map/mapping settings
 
-You can individually adjust the properties that Overviewer will use for each
-world by editing the world's `overviewer-settings.py` file. Properties here
-include the output path of the map (i.e. you can change this to your web
-server directory), and render settings. Please visit
-[their website](http://docs.overviewer.org/en/latest/config/) for information
-on config.
+If you get a `Permission denied` type of error, please see the [Troubleshooting section](#troubleshooting).
```
zanix: Remove "type of"

README.md
```diff
@@ -262,10 +257,13 @@ that you wish the world to use:
 
 ### Adjusting global server properties
 Default values in the script can be overridden by adding certain properties to one
-of the `mscs.defaults` files. The `mscs.defaults` files can be found found in one
+of the `mscs.defaults` files. The `mscs.defaults` files needs to be created in one
```
zanix: I wonder if we should have the installer create `/opt/mscs/mscs.defaults` and have the options commented out. That or include a file the user can copy and rename.

README.md
```diff
@@ -475,38 +475,69 @@ Where `<world>` is the name of the world you wish to map (omit for all
...
+Minecraft Server Control Script uses [Overviewer](http://docs.overviewer.org/en/latest/) to generate maps. 
+
+### Installing Overviewer
+
+Overviewer has pretty straightforward documentation to download and install the software:
+
+* [Debian/Ubuntu](http://overviewer.org/debian/info)
+* [RHEL/CentOS/Fedora](http://overviewer.org/rpms/info)
+
+*Once you follow the install page, come back here for further instructions.
```
zanix: Should we use blockquotes for notes like this? They stand out a little bit better when the markdown is rendered, but they do look odd when reading the raw file.

```
> __NOTE:__ Once you follow the install page...
```

> __NOTE:__ Once you follow the install page...

msctl
```diff
@@ -1394,7 +1394,19 @@ overviewer() {
   $OVERVIEWER_BIN --config=$SETTINGS_FILE >> $LOG_FILE 2>&1
   $OVERVIEWER_BIN --config=$SETTINGS_FILE --genpoi >> $LOG_FILE 2>&1
 
-  # Add the Google API key to the index.html. 
+  # Add the Google API key to the Overviewer index.html source files.
+  for f in $(find /usr/ -name "index.html" -print | grep overviewer_core/data/web_assets)
```
sandain: This command returns 2 symbolic links and one file on my server:
```
/usr/share/pyshared/overviewer_core/data/web_assets/index.html
/usr/lib/python2.6/dist-packages/overviewer_core/data/web_assets/index.html
/usr/lib/python2.7/dist-packages/overviewer_core/data/web_assets/index.html
```
Can we just edit the file directly rather than search for it?

msctl
```diff
@@ -1393,6 +1393,22 @@ overviewer() {
...
+                      sed -i "s/\?sensor=false/&\&key=$OVERVIEWER_KEY/g" "$f"
+                  fi
+          fi
+      fi
+  done
+
+  # Add the Google API key to the index.html in the maps folder. 
+  sed -i 's/\?sensor=false/&\&key='"$OVERVIEWER_KEY"'/g' "$MAPS_LOCATION/$1/index.html"
+
   # Announce the location to access the world map to players.
```
sandain: Do you need to edit both the source file and the generated file?  Do you need to edit both every time mapping is run?

zanix: I think only the source files needed to be edited.

README.md
sandain: Automatically generating a file for this location might not be a bad idea.  Either commented out (assuming the code can handle that, I think it does), or simply with the default values.

msctl
sandain: I imagine an update to Overviewer will overwrite any changes to the source file, so it would need to be checked.  Maybe the best option would be to just modify the generated file.. less messy.

zanix: That's right, we could override it with a custom assets option.
http://docs.overviewer.org/en/latest/config/?highlight=web_assets#custom-web-assets
>Any files in the custom web assets directory overwrite the default files.

Roflicide: >The problem with adding your key after the render [this is what I did in the msctl file] is that overviewer.py seems to copy the original index.html to your webroot and then perform the time consuming tile render. This means if you're waiting for overviewer.py to complete before running the sed to insert your key, and someone has gone exploring and built a backlog of tiles to render, your map may not work for a good chunk of time.

--> this happened to my server lol

> My solution is to add the key to /usr/share/pyshared/overviewer_core/data/web_assets/index.html after installing overviewer

^ So you're right in that we only need to modify the source file, the problem is I get a permission denied error when I try to do it since the user running the `msctl` script does not have permission

msctl
Roflicide: @sandain See my comment https://github.com/MinecraftServerControl/mscs/pull/160#discussion_r107267839

msctl
sandain: That is an even better option @zanix. I wonder if you can just have a replacement `index.html` file with the rest of the assets loaded from `/usr/share/pyshared/overviewer_core/data/web_assets/`.  The way the docs are worded, it sounds like it.

zanix: @sandain as I read the docs, that does seem how it works.

Roflicide: @zanix @sandain I tried this and it wouldn't render, I think you need all of the resources in the folder and not just the `index.html`

README.md
Roflicide: Thanks for noting, fixed

README.md
Roflicide: Thanks, done

README.md
Roflicide: Thanks!

README.md
Roflicide: updated, i think it looks pretty nice. thanks for the comment!

msctl
zanix: That kinda sucks, although we could copy the entire web_assets folder and then edit.

sandain:
What should we do about this pull request?  I haven't even looked at it in a while, but it looks like we never resolved some issues. Overviewer still hasn't switched away from the Google API, but they seem to be close.  Are there any changes in here that should be merged to master regardless of which API Overviewer is using?

Roflicide:
No, once they switch over to the other maps API no API keys will be required. The simplest solution would just be to wait it out; perhaps we could put a note in the documentation linking to this issue if anyone is curious and wants to use Overviewer before they update it.

zanix:
I like some of the documentation updates and auto creating a commented out `mscs.defaults` file. As for the API keys, I agree with @Roflicide, we could add some more documentation about the issue until it's resolved.

sandain:
I created pull request #166 to add the commented out `mscs.defaults` file. Let me know what you think in that pull request.

I'll need to look closer at this pull request to incorporate the changes to the documentation when I get another free moment. That is unless someone else has the time to create a new pull request with just that done.

PR #161, opened 2017-03-22
Fix missing quotation marks in json-query command

Roflicide:
Performing the `query-json` command would yield the first entry in the
json "players" array missing quotation marks; this fixes it

p.s. I couldn't figure out a way to fork the original project multiple times, so I just went and created a separate branch and then made the commit in there... if there's a better way to do this let me know b/c I'm still learning this stuff haha. Also sorry about committing directly to the master branch before

### DIFF
--- a/msctl
+++ b/msctl
@@ -1616,7 +1616,7 @@ queryDetailedStatusJSON() {
   if [ -n "$STATUS" ]; then
     NUMPLAYERS=$(printf "%s" "$STATUS" | cut -f 19)
     if [ $NUMPLAYERS -gt 0 ]; then
-      PLAYERS=$(printf "\"%s\"" "$STATUS" | cut -f 30)
+      PLAYERS=$(printf "%s\"" "$PLAYERS" $(printf "%s" "$STATUS" | cut -f $((30))))
       COUNTER=1
       while [ $COUNTER -lt $NUMPLAYERS ]; do
         PLAYERS=$(printf "%s, \"%s\"" "$PLAYERS" $(printf "%s" "$STATUS" | cut -f $((30+$COUNTER))))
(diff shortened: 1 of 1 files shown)


msctl
```diff
@@ -1616,7 +1616,7 @@ queryDetailedStatusJSON() {
   if [ -n "$STATUS" ]; then
     NUMPLAYERS=$(printf "%s" "$STATUS" | cut -f 19)
     if [ $NUMPLAYERS -gt 0 ]; then
-      PLAYERS=$(printf "\"%s\"" "$STATUS" | cut -f 30)
+      PLAYERS=$(printf "%s\"" "$PLAYERS" $(printf "%s" "$STATUS" | cut -f $((30))))
```
sandain: Ok, I think I understand the issue now.  Does this fix it?
```patch
diff --git a/msctl b/msctl
index 09a3548..dcd39d1 100755
--- a/msctl
+++ b/msctl
@@ -1584,7 +1584,7 @@ queryDetailedStatusJSON() {
   if [ -n "$STATUS" ]; then
     NUMPLAYERS=$(printf "%s" "$STATUS" | cut -f 19)
     if [ $NUMPLAYERS -gt 0 ]; then
-      PLAYERS=$(printf "\"%s\"" "$STATUS" | cut -f 30)
+      PLAYERS=$(printf "\"%s\"" $(printf "%s" "$STATUS" | cut -f 30))
       COUNTER=1
       while [ $COUNTER -lt $NUMPLAYERS ]; do
         PLAYERS=$(printf "%s, \"%s\"" "$PLAYERS" $(printf "%s" "$STATUS" | cut -f $((30+$COUNTER))))

```

Roflicide: Yea that fixed it--thanks!

Issue #164, opened 2017-06-07
Server update is bypassed with no user notification if previous update attempted in less that 1440 minutes prior (by default)


NetworkEng:
Running mscs/msctl update will not update to latest minecraft server version 1.12. Latest version updated (on my host) is 1.11.2.  Apologies for the quick issue create with no additional troubleshooting, but will attempt to do that tonight when done with work.

sandain:
Huh, it worked for me.  Try removing your `version_manifest.json` file, or at least check to make sure that 1.12 is listed as the latest release version in the file.

NetworkEng:
Thanks Jason.  I checked /opt/mscs/version_manifest.json and it has a current timestamp (of when I manually ran the 'mscs update'), but the file was not the same as on the web at https://launchermeta.mojang.com/mc/game/version_manifest.json.

Found that this line:
`547:    find "$VERSIONS_JSON" -mmin +"$VERSIONS_DURATION" -delete
`
Along with this conditional:
`550:  if [ ! -s $VERSIONS_JSON ]; then
`
`551:    $WGET --no-use-server-timestamps -qO "$VERSIONS_JSON" "$MINECRAFT_VERSIONS_URL"
`
To be the culprit.  As you already know, it won't update if that version_manifest.json file already exists.  It seems the update was published some time this morning, after my daily cron job to update had run.

Yes, one can manually configure each world VERSIONS_DURATION to be less than one day for each world, but (to me) this is unexpected behavior.  I'd think that most users would expect a manual run of update to succeed if there is a newer version available, and not be bypassed if the update had been run in less than 1440 minutes.  For me, I wasn't aware that this restriction existed by default.

Perhaps this issue should be a feature request to consider changing the default behavior, or, add a command line option to "force" an update, bypassing the VERSIONS_DURATION limitation.

Thanks for your consideration.

-Matt

NetworkEng:
In addition to a "force" command line option, would probably be good to inform the user that it did not attempt to update due to the last attempted update being less than the VERSIONS_DURATION timeframe, and to use the "force" option to bypass it.

sandain:
Perhaps we should lower the VERSIONS_DURATION variable to something like 30 minutes -- a day is a long time and 30 minutes would do the job (10 minutes would probably be more than enough too).

[...]

PR #165, opened 2017-06-13
RFC: Rework version_manifest.json update function

sandain:
RFC: Rework the `version_manifest.json` update function so that we only try to download the file once each time the script is executed. Prior to this change the update code would run multiples times per execution of the script -- subsequent downloads would be skipped if the file was new enough (defined by VERSIONS_DURATION), but was a hacky solution.

Not all command line options require the version_manifest.json file to exist, so we might want to move the call to the new update function `updateVersionsJSON` elsewhere.

I also reduced the default duration of this file to 30 minutes.

This does not cover all of the requests in Issue #164, but it is a start.

### DIFF
--- a/msctl
+++ b/msctl
@@ -1869,7 +1869,7 @@ fi
 #   mscs-worlds-location=/opt/mscs/worlds
 #   mscs-versions-url=https://launchermeta.mojang.com/mc/game/version_manifest.json
 #   mscs-versions-json=/opt/mscs/version_manifest.json
-#   mscs-versions-duration=1440
+#   mscs-versions-duration=30
 #   mscs-lockfile-duration=1440
 #   mscs-default-world=world
 #   mscs-default-port=25565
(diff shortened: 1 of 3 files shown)


zanix:
Could we use a global variable that is set to `TRUE` when `version_manifest.json` is updated? Then subsequent runs of `updateVersionsJSON` can check this variable and the last update time and skip if needed.

zanix:
Wait, I see that `updateVersionsJSON` is now only ran once, nevermind my suggestion.

msctl
```diff
@@ -2030,6 +2034,10 @@ OVERVIEWER_URL=$(getDefaultsValue 'mscs-overviewer-url' 'http://overviewer.org')
 MAPS_LOCATION=$(getDefaultsValue 'mscs-maps-location' $LOCATION'/maps')
 MAPS_URL=$(getDefaultsValue 'mscs-maps-url' 'http://minecraft.server.com/maps')
 
+# Grab the latest version information.
+# ---------------------------------------------------------------------------
+updateVersionsJSON
+
```
sandain: I just audited the options that require that the `version_manifest.json` be up to date, and I narrowed it down to _start_, _update_, and _overviewer_.  We should move this down into those options to further reduce the need to run this function.

msctl
```diff
@@ -2483,7 +2504,16 @@ case "$1" in
...
     fi
   ;;
-  update)
+  update|force-update)
+    # Handle the force option.
+    if [ "$1" = "force-update" ]; then
+      if [ -s $VERSIONS_JSON ]; then
+        printf "Removing cached version manifest.\n"
+      fi
+      rm -f $VERSIONS_JSON
```
zanix: Should we handle this the same way `updateVersionsJSON` does by renaming `version_manifest.json` to `version_manifest.json.bak`

sandain: That sounds like a good idea. The only reason we wouldn't want to rename the manifest is if it has a size of 0 and fails the -s test. In this case, the proper thing to do is delete the file.  Adding a copy command to the _if_ block seems like it would do the job.

sandain:
I'm going to go ahead and merge these changes.  I've tested it fairly well and don't see any new issues.  Hopefully I didn't miss something.

zanix:
👍

PR #166, opened 2017-06-29
Create an example mscs.defaults file if it is missing

sandain:


msctl
```diff
@@ -1825,6 +1956,10 @@ elif [ -r "$HOME/mscs.defaults" ]; then
 elif [ -r "$HOME/.config/mscs/mscs.defaults" ]; then
   MSCS_DEFAULTS="$HOME/.config/mscs/mscs.defaults"
 fi
+if [ ! -r "$MSCS_DEFAULTS" ]; then
```
sandain: This should be `-s` (True if file exists and has a size greater than zero) rather than `-r` (True if file exists and is readable).

sandain:
Any feedback here? This is a pretty simple change.

zanix:
Looks alright to me.

Roflicide:
Yes, this looks great!

PR #168, opened 2017-06-30
Update LICENSE

Roflicide:
it's that time of year again


sandain:
Thanks. I always forget to do this.

PR #173, opened 2017-08-25
Update readme

Roflicide:
Clarified backup + mapping procedures and added note on Java 8 requirement as of Minecraft v1.12 (now required)


sandain:
Looks good to me. Thanks.

Issue #174, opened 2017-08-25
Give better error when Java version is wrong


sandain:
According to @Roflicide, Java 8 is required for Minecraft versions >= 1.12 (see #173).  Are there any versions of Minecraft that can work with Java 6, and if so when did 7 become required?

Anyway, we have code in the script for testing Minecraft versions and we can extract the Java version with:
```
java -version 2>&1 | perl -ne 'if ($_ =~ /version\s"([\d\.\_]+)"/) { print $1 }'
```
All we need to do is setup a trap for invalid combinations and produce a better error message.

See #171 for the inspiration behind this.

sandain:
According to [Mojang](https://help.mojang.com/customer/en/portal/articles/2636196-minecraft---java-6-and-7-support), up until 1.12 both Java 6 and 7 were supported.  Java 5 support was killed off with [version 1.6](https://minecraft.gamepedia.com/1.6.1) if we want to go that far back in time.

zanix:
Nice, I was attempting to find this info as well.
Maybe check the Java version against the Minecraft version and display an error?
Also, a server can potentially have multiple versions of Java installed using `update-alternatives`, do we want to handle this or just keep using whatever `which java` returns?

sandain:
Ohh goodness, I hadn't thought of that.  Debian based systems use `update-alternatives`, but I don't think that is universal.  The heuristics for this would be tricky.  Maybe we should just rely on `which java` and require the usage of the `mscs.defaults` or `mscs.properties` files for special cases.

sandain:
Here is another thing to note: servers below Minecraft version 1.3.1 are not available at the update URL that we use.

sandain:
Here are a couple of things to note: servers below Minecraft version 1.3.1 are not available at the update URL that we use, and a 1.3.1 server and client works just fine with Java 8.

Roflicide:
I agree with @zanix. I think all we really need is something that checks the java version and tells the user the supported Minecraft versions based on the JRE they have (from `which java`) if they are running a JRE less than java 8. So something like:

If their running Java 5: Minecraft versions 1.3.1-1.6 supported
If their running Java 6 & 7: Minecraft versions 1.6.1-1.11.2 supported

We could add it so that they could change the `mscs.defaults` or `mscs.properties` files to specify JRE location, but I don't think it's really necessary. Maybe if someone specifically requests that then we could do it.

sandain:
[...]

PR #176, opened 2017-09-12
Give better error when Java version is wrong

sandain:
See Issue #174

### DIFF
--- a/msctl
+++ b/msctl
@@ -41,7 +41,6 @@ PROG=$(basename $0)
 # Required Software
 # ---------------------------------------------------------------------------
 # Detect its presence and location for later.
-JAVA=$(which java)
 PERL=$(which perl)
 PYTHON=$(which python)
 WGET=$(which wget)
@@ -973,6 +972,7 @@ getServerCommand() {
   local CURRENT_VERSION SERVER_VERSION SERVER_JAR SERVER_LOCATION
   local SERVER_ARGS INITIAL_MEMORY MAXIMUM_MEMORY
   local DEFAULT_ARGS DEFAULT_INIT_MEM DEFAULT_MAX_MEM DEFAULT_COMMAND
+  local DEFAULT_JAVAVM JAVA JAVA_VERSION VERSION_TEST
   CURRENT_VERSION=$(getCurrentMinecraftVersion "$1")
   if [ $? -ne 0 ]; then
     printf "$CURRENT_VERSION\n"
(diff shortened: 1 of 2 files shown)


msctl
```diff
@@ -993,6 +993,54 @@ getServerCommand() {
...
+    1.9)
+      # Java 9 supports all server versions.
+    ;;
+    1.8)
+      # Java 8 supports all server versions.
+    ;;
+    1.7)
+      # Support for Java 7 ended with version 1.12.
+      VERSION_TEST=$(compareMinecraftVersions $SERVER_VERSION 1.12.1)
+      if [ $VERSION_TEST -ge 0 ]; then
```
sandain: This should be:
```
if [ $VERSION_TEST -gt 0 ]; then
```

I was just wondering about snapshot versions. Do all 1.12 snapshot versions support Java 6 and 7 but only the final release, 1.12.1 drop support?  Maybe it would be best to assume that snapshot versions have the same requirements as the release versions.  In that case, maybe we should do the following comparison:
```
VERSION_TEST=$(compareMinecraftVersions $SERVER_VERSION 1.11.2)
if [ $VERSION_TEST -le 0 ]; then
```

msctl
```diff
@@ -993,6 +993,54 @@ getServerCommand() {
...
+        echo "Error: Server version $SERVER_VERSION not supported by Java 7."
+        exit 1
+      fi
+    ;;
+    1.6)
+      # Support for Java 6 ended with version 1.12.
+      VERSION_TEST=$(compareMinecraftVersions $SERVER_VERSION 1.12.1)
+      if [ $VERSION_TEST -ge 0 ]; then
```
sandain: Same as above.

msctl
```diff
@@ -993,6 +993,54 @@ getServerCommand() {
...
+        echo "Error: Server version $SERVER_VERSION not supported by Java 6."
+        exit 1
+      fi
+    ;;
+    1.5)
+      # Support for Java 5 ended with version 1.6.
+      VERSION_TEST=$(compareMinecraftVersions $SERVER_VERSION 1.6.1)
+      if [ $VERSION_TEST -ge 0 ]; then
```
sandain: Same as above. Last 1.5 release was 1.5.2, but this is less of an issue because snapshots are only avail for the 1.12 series.

msctl
Roflicide: I think you're right. MC Wiki says Release 1.12 is the first version to use/require Java 8, so I think it's safe to assume that the snapshots (which are < Release 1.12) are probably Java 6/7 compatible.

Roflicide:
@sandain 
> This change assumes that Java is being used for all server types. Are there servers that are written in >other languages and thus provide their own binary in use? I think this change would break those >configurations (i.e., those that change the mscs-server-command variable to run a different binary than >java). I'm not sure what the proper fix would be.

The amount of people utilizing a different `mscs-server-command` variable must be incredibly small. In my opinion, I would just add a check that if their using something different than good ol' java (as specified in the  `mscs-server-command` variable) to disable Java version checking and then prompt them a message that Java version checking is disabled

sandain:
I was kind of thinking the same thing.  I'll take another look at the code soon and see if I can come up with something simple that can handle this situation.

msctl
sandain: Actually, according to the [MC wiki](https://minecraft.gamepedia.com/17w13a), 17w13a (the 2nd snapshot) is the first version to require Java 8. Since I don't want to hard code snapshot versions into the script, I think the modified test I posted (<= 1.11.2) would be the most appropriate.

sandain:
Well I obviously never got around to working on this pull request.  I'm going to go ahead and just close it as I don't see any need to support older Java versions (it is a security issue to run old versions anyway).

Issue #181, opened 2017-12-15
Overviewer Setup


shawnbyday:
The overviewer setup seems a little bit out of date. Not due to mscs but due to some changes with Overviewer. Specifically with the GoogleMapsAPI key.

sandain:
That doesn't surprise me. They were working on a new version that didn't require GoogleMaps if I recall correctly.  If you feel up to the task, I'll review patches to the README if that documentation needs updated.  The wiki should be editable by everyone if you want to touch the documentation there.  Otherwise, I'll take a look at this when I have some free time.

shawnbyday:
You can find the following "work around" by @supernerdpp below:

I found this on the overviewer github issues page: [link](https://github.com/overviewer/Minecraft-Overviewer/issues/1308)

Hopefully this helps someone!

---

You have to get a key now, but anyone can get one from: https://developers.google.com/maps/documentation/javascript/get-api-key

Then in the index.html file, you have to make the following modification:
Change Line 14:
<script type="text/javascript" src="https://maps.google.com/maps/api/js?sensor=false"></script>

to:
<script type="text/javascript" src="https://maps.google.com/maps/api/js?key=<your key here>&sensor=false"></script>

Omitting <>

shawnbyday:
I may take a look at it as well. I'm always shy of changing documentation because I'm never sure if I'm 100% correct or not. Fear of failure!

sandain:
Fixed with patch by @Roflicide

PR #183, opened 2018-03-21
Readme overviewer update #2

Roflicide:
Addendum to issue [#181](https://github.com/MinecraftServerControl/mscs/issues/181). 

Turns out that you actually need to edit in the API key into multiple points in your system where the source of overviewer is installed. Otherwise, the `index.html` for your world will be overwritten without including that API key. (Tested this out myself on my server)


sandain:
Looks good. Thanks!

Issue #187, opened 2018-05-22
Corrupt jar File with mscs update


HappyWarbird:
mscs force-update downloads corrupt jar files since end of April, looks like amazonaws changed something. Here's the list of /server folder
`-rw-r--r-- 1 minecraft minecraft 31534189 Apr  5 16:56 minecraft_server.18w14b.jar`
`-rw-r--r-- 1 minecraft minecraft 31583033 Apr 12 11:00 minecraft_server.18w15a.jar`
`-rw-r--r-- 1 minecraft minecraft 31620910 Apr 20 15:59 minecraft_server.18w16a.jar`
`-rw-r--r-- 1 minecraft minecraft        0 May 15 06:00 minecraft_server.18w19b.jar`
`-rw-r--r-- 1 minecraft minecraft        0 May 16 06:00 minecraft_server.18w20a.jar`
`-rw-r--r-- 1 minecraft minecraft        0 May 17 06:00 minecraft_server.18w20b.jar`
Manual DL from minecraft.net is working normal.

Entry in mscs.defaults
`https://s3.amazonaws.com/Minecraft.Download/versions/$SERVER_VERSION/minecraft_server.$SERVER_VERSION.jar`
Replaced $SERVER_VERSION with the Snapshot 18w20c occurs in AccessDenied on amazonaws.
Replaced $SERVER_VERSION with Release 1.12.2 works.

Jonicho:
I think i found the issue:

`http://s3.amazonaws.com/Minecraft.Download/versions/versions.json`
says
`__comment": "This URL is being phased out! Please update your scripts to check https://launchermeta.mojang.com/mc/game/version_manifest.json instead. Thanks <3 --Dinnerbone"`

The download links in `https://launchermeta.mojang.com/mc/game/version_manifest.json` work.

So mscs may need an update to continue supporting the latest minecraft versions.

I would help here, but i'm not that familiar with bash scripting.

zanix:
I think the code was switched to use `https://launchermeta.mojang.com/mc/game/version_manifest.json` a long time ago. 
Although the client download URL still points to `mscs-default-client-url=https://s3.amazonaws.com/Minecraft.Download/versions/\$CLIENT_VERSION/\$CLIENT_VERSION.jar`
And the server download URL points to `mscs-default-server-url=https://s3.amazonaws.com/Minecraft.Download/versions/\$SERVER_VERSION/minecraft_server.\$SERVER_VERSION.jar`

[...]

PR #189, opened 2018-05-25
msctl: fix download of server and client

sandain:
This should fix issue #187.

### DIFF
--- a/msctl
+++ b/msctl
@@ -821,7 +821,7 @@ getClientLocation() {
 # @return CLIENT_URL
 # ---------------------------------------------------------------------------
 getClientURL() {
-  local CURRENT_VERSION CLIENT_VERSION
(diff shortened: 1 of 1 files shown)


zanix:
I was just looking at the JSON and noticed there is also a sha1 signature at `$json->{downloads}{client}{sha1}`. A good feature would be to run `sha1sum` on the file after it downloaded and compare it to the JSON, that is, if the download comes from the versions JSON file.

msctl
```diff
@@ -1411,6 +1445,9 @@ updateClientSoftware() {
     printf "$CLIENT_URL\n"
     exit 1
   fi
+echo $CLIENT_JAR
+echo $CLIENT_LOCATION
+echo $CLIENT_URL
```
sandain: Oops, I left some debugging code in place.

PR #192, opened 2018-06-21
Fixed README documentation of overviewer, because it switched to leaflet

ssummer1337:
Overviewer has switched to leaflet which is an open-source javascript map library.

Therefore it is not necessary to add a Google API key to the index.html.


sandain:
Thanks for the patch! I was wondering when they were going to finish the switch to Leaflet.

PR #202, opened 2019-01-24
Atom format shell formatting

Roflicide:
Applies the Atom `format-shell` formatting (using the default 2 spaces formatting in the editor)

### DIFF
--- a/msctl
+++ b/msctl
@@ -2024,38 +2024,38 @@ fi
 # ---------------------------------------------------------------------------
 while [ "$1" != "${1#-}" ]; do
   case "$1" in
-  -p)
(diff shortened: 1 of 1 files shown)


sandain:
Everything looks ok, but the last part of this patch reorders the `case` statement.  I'll need to spend more time looking at this then I have right now.  I'll try to spend some time this evening reviewing this patch.

Roflicide:
I just realized there is an option to format switch case statements. how does the last revision look?

sandain:
This looks great.  I'll go ahead and merge this patch.  Thanks for making this easy on me.

Issue #185, opened 2018-04-20
Multiple world commands


csdt:
Many commands can deal with multiple worlds at the same time:
```
* start [world]
* stop [world]
* force-stop [world]
* restart [world]
* force-restart [world]
* status [world]
* logrotate [world]
* backup [world]
* map [world]
* update [world]
```

But we can only either execute such a command for one single world, or all of them.
I think it would be great to be able to specify multiple worlds for those commands

sandain:
This would be fairly easy to implement if someone is looking for a project. I would, but I don't really have the time right now.

sandain:
@Roflicide sent in PR #205 to add this feature, so closing this issue.  Thanks @Roflicide!

PR #203, opened 2019-01-24
Make commands accept multiple world arguments

Roflicide:
Pull request in reference to Issue #185  -- makes so applicable commands can accept more than 1 world argument

suggestions / comments welcome as im new to shell scripting =)

### DIFF
--- a/msctl
+++ b/msctl
@@ -2303,7 +2303,9 @@ MAPS_URL=$(getDefaultsValue 'mscs-maps-url' 'http://minecraft.server.com/maps')
 
 # Respond to the command line arguments.
 # ---------------------------------------------------------------------------
-case "$1" in
+# Stores the user-issued command.
+command=$1
+case "$command" in
   start)
     # Grab the latest version information.
     updateVersionsJSON
@@ -2318,14 +2320,17 @@ case "$1" in
       echo "Or create the default world at the default port:"
       echo "  $PROG create"
       exit 1
-    elif [ -n "$2" ]; then
-      if isWorldEnabled "$2"; then
-        WORLDS="$2"
-      elif [ -n "$2" ]; then
-        printf "World '$2' not recognized.\n"
-        printf "  Usage:  $PROG $1 <world>\n"
-        exit 1
-      fi
+    elif [ "$#" -ge 2 ]; then
+      shift
+      for arg in "$@"; do
+        if isWorldEnabled "$arg"; then
+          WORLDS="$WORLDS $arg"
+        else
+          printf "World '$arg' not recognized.\n"
+          printf "  Usage:  $PROG $command <world>\n"
+          exit 1
+        fi
+      done
     fi
     # Start each world requested, if not already running.
     printf "Starting Minecraft Server:"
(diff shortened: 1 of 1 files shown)


msctl
```diff
@@ -2303,7 +2303,9 @@ MAPS_URL=$(getDefaultsValue 'mscs-maps-url' 'http://minecraft.server.com/maps')
 
 # Respond to the command line arguments.
 # ---------------------------------------------------------------------------
-case "$1" in
+# Stores the user-issued command.
+command=$1
```
sandain: you could add a `shift` command here since you are storing `$1` in `$command`.  This would save you from having to use `shift` in each section below.

msctl
```diff
@@ -2318,14 +2320,17 @@ case "$1" in
...
-    elif [ -n "$2" ]; then
-      if isWorldEnabled "$2"; then
-        WORLDS="$2"
-      elif [ -n "$2" ]; then
-        printf "World '$2' not recognized.\n"
-        printf "  Usage:  $PROG $1 <world>\n"
-        exit 1
-      fi
+    elif [ "$#" -ge 2 ]; then
+      shift
```
sandain: This would then change to
```
+    elif [ "$#" -ge 1 ]; then
```

msctl
sandain: Also, every other variable in the script is in all caps. To match the rest of the script, it should be:
```
COMMAND=$1
```
This is just esthetics

msctl
```diff
@@ -2303,7 +2303,9 @@ MAPS_URL=$(getDefaultsValue 'mscs-maps-url' 'http://minecraft.server.com/maps')
 
 # Respond to the command line arguments.
 # ---------------------------------------------------------------------------
-case "$1" in
+# Stores the user-issued command.
+command=$1
+case "$command" in
```
sandain: This would then be:
```
case "$COMMAND" in
```

msctl
Roflicide: thanks for the comments :) 

> you could add a `shift` command here since you are storing `$1` in `$command`. This would save you from having to use `shift` in each section below.

the reason i didn't do this is because there are other parts of the script where they only accept one argument, so I'd have to shift all of those arguments down by 1. for instance, right now to create a world it calls `createWorld $2 $3 $4`. if I had the shift command applied globally, then I'd have to modify that statement so it's `createWorld $1 $2 $3`, and do this with several other statements in the file. Do you think its better to shift the parameters globally and just make these tweaks, or keep it as it is right now (where it's only shifted in certain code blocks)? 

> Also, every other variable in the script is in all caps. To match the rest of the script, it should be:
> 
> ```
> COMMAND=$1
> ```
> This is just esthetics

okay, i wasn't sure if it was a style thing or if the caps actually denoted a constant or something

sandain: I'm fine leaving this patch as is, although it does need to be updated to resolve the conflict created by not sorting the `case` statement options.

msctl
Roflicide: fixed

msctl
Roflicide: i think i resolved all the issues with the latest commit =)

sandain:
@Roflicide, this branch still has merge conflicts stopping me from merging.  It also has a bunch of whitespace/etc changes that your Atom format patch either missed or are duplicated here.  Can you create a new pull request based on master with just the changes you had in mind?

PR #205, opened 2019-03-20
Make commands accept multiple world arguments

Roflicide:
Updated pull request (with hopefully no merge issues) for Issue #185

PR #207, opened 2019-03-20
Make more commands accept multiple world arguments + update doc

Roflicide:
This is the second part to pull request #205 for adding commands that accept multiple world arguments. There were a few commands that I had not changed in that pull request to accept multiple world arguments (like `mscs enable` and `mscs disable` ) so this pull request adds them.

This pull request also updates the documentation in both the script and in the README.

The complete, updated list of multi-world commands is below. If no arguments are given it will perform the command on all the worlds: 

````
start
stop
force-stop
restart
force-restart
disable
enable
status
sync
logrotate
backup
map
update
force-update
query
````

These commands remain single-world or single-argument commands:

````
create
delete
ls
list
broadcast
send
console
watch
list-backups
restore-backup
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,125 +547,115 @@ user or as `root` (through sudo).
 Note: If the script is run as the `root` user, all important server processes
 will be started using the `minecraft` user instead for security purposes.
 
-    sudo mscs [option]
 
 ````
-* start [world]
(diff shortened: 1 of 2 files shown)


sandain:
Thanks for your work on this @Roflicide.  Sorry for the slow merge

PR #209, opened 2019-04-12
Add commands for importing and renaming worlds

sandain:


sandain:
I've had this code sitting on my system for a long time.  It seems like it could be useful.

I would just like to get at least one other person to test it before I merge it.

PR #212, opened 2019-05-05
make mscs status player output consistent

conrad784:
used to display first name with " " around it, but any additional name would be displayed without


sandain:
I'm not sure why those quotes are there.. Probably a copy-paste issue.  Thanks for the patch!

PR #213, opened 2019-05-11
Update import script to work with dash 0.5.8-2.4

avanheuvelen:
Tested on Debian 9.

I haven't tested if this works with bash, but I think it should. Would be nice if someone could test it first.


avanheuvelen:
Never mind, just found out I should use mscs instead of msctl

sandain:
I didn't realize that I had used a bashism here.  Thanks for the patch!

Issue #214, opened 2019-05-16
Error Updating to 1.14


strmtrupr2:
Hi, I saw some other issues and this is what I've done so far:
* git pull, make update
* ran this code:
`perl -0777ne '
    use JSON;
    use LWP::Simple;
    my $json = decode_json ($_);
    my $version;
    foreach $ver (@{$json->{versions}}) {
      $version = $ver if ($ver->{id} eq "1.13.1");
    }
    $json = decode_json (get ($version->{url}));
    print $json->{downloads}{server}{url};
  ' /opt/mscs/version_manifest.json`
This worked fine.
* Then, I ran this code:
`perl -0777ne '
    use JSON;
    use LWP::Simple;
    my $json = decode_json ($_);
    my $version;
    foreach $ver (@{$json->{versions}}) {
      $version = $ver if ($ver->{id} eq "1.14");
    }
    $json = decode_json (get ($version->{url}));
    print $json->{downloads}{server}{url};
  ' /opt/mscs/version_manifest.json`
and it errored with this message:
> malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "(end of string)") at -e line 9.

Any ideas?

Thanks,

Scott

sandain:
Hi Scott.

That's weird. I just checked the [version manifest](https://launchermeta.mojang.com/mc/game/version_manifest.json), and `1.14` is a valid version.  I'll have to dig more into this when I get the chance, probably not until this weekend.. Does version `1.14.1` work?

strmtrupr2:
No, tried 1.14.1 too...

Scott

On May 16, 2019, at 1:58 AM, Jason M. Wood <notifications@github.com> wrote:

Hi Scott.

That's weird. I just checked the version manifest, and 1.14 is a valid version. I'll have to dig more into this when I get the chance, probably not until this weekend.. Does version 1.14.1 work?

sandain:
I just checked if I could update my server, and I had no problem.  I modified one of my servers to run `1.14`, another is set to use the latest release version, and I also have one set to run the latest snapshot.  I'm not sure what the issue could be yet...
```
$ mscs status
Minecraft Server Status:
  alpha: running version 1.14 (0 of 20 users online).
    Process ID: 8722.
    Memory used: 626232 kB.
  beta: running version 1.14.1 (0 of 20 users online).
    Process ID: 8938.
    Memory used: 917272 kB.
  lobby: running.
    Process ID: 9143.
[...]

PR #215, opened 2019-05-19
package dependencies for fedora

conrad784:
Those should be all dependencies necessary to run mscs on Fedora.
See #214


sandain:
Thanks for the patch @conrad784.

Issue #218, opened 2019-06-04
Imported world getting regenerated


ObiWanCanOweMe:
I'm trying to import a world I created (v1.14.2) using the `itzg/minecraft-server` docker image.  Works just fine using Docker and imports into mscs without issue.  However, when I start the world for the first time it always generates a new map.

Steps to reproduce:
1. Copy world directory to new server with mscs installed.
2. Run the following commands:

`# mscs import /home/user/obi-world obi-world 25565`
`# mscs start obi-world`

Here is an excerpt from the console after starting the world for the first time:
```
# mscs console obi-world
Connecting to server console: obi-world.
  Hit <Ctrl-D> to detach.
[15:37:16] [main/WARN]: Ambiguity between arguments [teleport, destination] and [teleport, targets] with inputs: [Player, 0123, @e, dd12be42-52a9-4a91-a8a1-11c01849e498]
[15:37:16] [main/WARN]: Ambiguity between arguments [teleport, location] and [teleport, destination] with inputs: [0.1 -0.5 .9, 0 0 0]
[15:37:16] [main/WARN]: Ambiguity between arguments [teleport, location] and [teleport, targets] with inputs: [0.1 -0.5 .9, 0 0 0]
[15:37:16] [main/WARN]: Ambiguity between arguments [teleport, targets] and [teleport, destination] with inputs: [Player, 0123, dd12be42-52a9-4a91-a8a1-11c01849e498]
[15:37:16] [main/WARN]: Ambiguity between arguments [teleport, targets, location] and [teleport, targets, destination] with inputs: [0.1 -0.5 .9, 0 0 0]
[15:37:16] [Server thread/INFO]: Starting minecraft server version 1.14.2
[15:37:16] [Server thread/INFO]: Loading properties
[15:37:16] [Server thread/INFO]: Default game type: SURVIVAL
[15:37:16] [Server thread/INFO]: Generating keypair
[15:37:17] [Server thread/INFO]: Starting Minecraft server on *:25565
[15:37:17] [Server thread/INFO]: Using epoll channel type
[15:37:17] [Server thread/INFO]: Preparing level "obi-world"
[15:37:17] [Server thread/INFO]: Found new data pack vanilla, loading it automatically
[15:37:17] [Server thread/INFO]: Reloading ResourceManager: Default
[15:37:20] [Server thread/INFO]: Loaded 6 recipes
[15:37:23] [Server thread/INFO]: Loaded 811 advancements
[...]

PR #219, opened 2019-06-05
Fix world imports

sandain:
See issue #218

### DIFF
--- a/msctl
+++ b/msctl
@@ -456,6 +456,24 @@ enableWorld() {
 }
 
 # ---------------------------------------------------------------------------
+# Import a world.
+#
+# @param 1 The location of the world to import.
+# @param 2 The name of the new imported world.
+# @param 3 The port of the new imported world.
+# @param 4 The IP address of the new imported world.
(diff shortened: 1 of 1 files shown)


msctl
```diff
@@ -455,6 +455,24 @@ enableWorld() {
...
+# @param 3 The port of the new imported world.
+# @param 4 The IP address of the new imported world.
+# ---------------------------------------------------------------------------
+importWorld() {
+  local NAME MCUSER MCGROUP
+  mkdir -p "$WORLDS_LOCATION/$2"
+  cp -R $1/* $WORLDS_LOCATION/$2
+  chown $(id -un):$(id -gn) -R $WORLDS_LOCATION/$2
+  NAME=$(getServerPropertiesValue $2 'level-name')
+  mv $WORLDS_LOCATION/$2/$NAME $WORLDS_LOCATION/$2/$2
```
zanix: Might want to add an exists check before calling `mv`

sandain: That is a good idea.  I just committed the change to the PR

PR #222, opened 2019-08-24
Fix overviewer mapping when world mirroring is enabled

Wicloz:
When a backup is made of a server with world mirroring enabled, a '-original' postfix is added to the world name, causing mapping to fail. Adding this postfix to the config manually will not solve the issue, as backups made when the server is off will not include this postfix. Detecting the appropriate world when mapping starts will fix the issue, and will not affect servers without world mirroring enabled.

### DIFF
--- a/msctl
+++ b/msctl
@@ -1630,9 +1630,9 @@ overviewer() {
   updateClientSoftware "$1"
   # Create a default Overviewer settings file if it is missing.
   if [ ! -e $SETTINGS_FILE ]; then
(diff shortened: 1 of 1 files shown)


sandain:
Hi @Wicloz.  Sorry it took me so long to get back to you on this PR.  It looks like a reasonable change.  I assume that you have tested this and that it fixes the problem, so I'll go ahead and accept the change.

Thanks for the patch.

PR #228, opened 2019-11-13
add mscs-jvm-args

rickyninja:


rickyninja:
Hey @sandain I added some tests too.  Please let me know if there are any issues with how tests were implemented.

zanix:
I've never seen tests for bash/sh before. Neat!

sandain:
I tried to setup tests for sh before, but didn't like what I had done so I abandoned it.  This looks like a nice clean method.  Good work!

I was a little confused about the change you made to printf (addition of `--`) until I did a little research.  I learned something new!

I don't have time to test the changes in this PR, but I did look over the code closely and it looks correct.  Thank you for working on this @rickyninja.

PR #229, opened 2019-11-15
fix test for getServerCommand

rickyninja:
I noticed an error in this test after the prior PR was already closed.

```
jeremys@skynet> printf $got
/usr/bin/java

jeremys@skynet> printf "$got"
/usr/bin/java -Xms128M -Xmx2048M -Dtesting-the-jvmargs -jar /home/jeremys/mscs/server/minecraft_server.fakedVersion.jar nogui
```


sandain:
Looks good. Thanks for catching the error and sending the patch.

Issue #232, opened 2019-12-11
query server shuts down if queried too early


sheimer:
if i request a status right after i started a world, the query server quits and does not start again.

i changed "msctl", lines 1814 to 1815 are added:

```
1808   if [ -n "$RESPONSE" ]; then
1809     # Return the response in the format requested.
1810     $PERL -e '
1811       $packed = join "", map { pack ("C", hex($_)) } ("'$RESPONSE'" =~ /(..)/g);
1812       printf "%s\n", join "\t", unpack ("'$4'", $packed);
1813     '
1814   else
1815     ps a | grep socat | grep "$1" | grep tail >/dev/null || queryStart "$1"
1816   fi
```

with that change i dont have to restart the whole world again to get the query server up and running again, and after one "restart by additional queryStart" i can immediately use query or status again, without the server shutting down.

i intend to request the status right after world starting, since i write a web app to control the mscs script, so it would be very nice to have this extra lines in the repository :)

sorry for writing this as an issue, i dont know how to open a merge request...if you need more information about the query server shutting down in the first place, feel free to ask!

kind regards, 

swen

sandain:
Hi @sheimer.  Thanks for the patch!  I'll have to find some time to give this a test

sandain:
I finally found some time to test this patch. It seems to work as expected, so I accepted the merge request.  Sorry for taking so long to get to the review. Thanks for the patch.

Roflicide:
What's the purpose of having an OR operator with `ps a | grep socat | grep "$1" | grep tail >/dev/null` ? Why not just take it out and simply have `queryStart $1` ?

sheimer:
there is a timeframe where the queryserver is starting, but still not responding. so in that timeframe you will get to the else part of the condition, and the part before the "or" prevents the query server to be started multiple times.

Roflicide:
@sheimer Thanks, that makes sense! I think I realized why this issue was occurring in the first place after some debugging: the query port hasn't opened yet so the socat connection automatically drops. Your solution of reconnecting in the `querySendPacket` function definitely fixes the issue, but it's downstream from the initial issue and takes a bit of time and a few `running (query server offline)` responses before the query server gets back up. I'm going to try and fix the source issue (waiting until the query port opens before connecting socat) so we can get a quicker/more robust response

Issue #169, opened 2017-07-13
Query server frequently offline


jwilling:
Hey! First of all, thanks to all of the contributors for the work done on this project. It's been extremely helpful and has saved me quite a lot of time.

I'm having an issue with the status commands through mscs. My server has two worlds running on non-standard ports. They both work just fine, and can be queried using an online tool with the expected response. When I try to query using `mscs status`, however, I tend to get this after the server has been running for about an hour (with a couple users online):

```
$ mscs status
Minecraft Server Status:
  wookies: running (query server offline).
    Process ID: 17617.
    Memory used: 1338836 kB.
  wookiescreative: running version 1.12 (0 of 20 users online).
    Process ID: 17816.
    Memory used: 1260216 kB.
```

Any idea what's causing this?

sandain:
Hi @jwilling.  I'm not sure what is going on here.  It sounds like you have tried everything I would recommend.

Here is how it works:
The script starts a `tail` process ([line 1595](https://github.com/MinecraftServerControl/mscs/blob/master/msctl#L1595)) that pipes anything that ends up in the `query.in` file to the Query server via `socat`. Responses from the Query server via `socat` are piped into the `query.out` file.  If the `tail` or `socat` process have issues for some reason, the Query server will be listed as offline.  Perhaps you have a permission issue on one of the two in/out files?  I don't know.

Roflicide:
Hey @jwilling,

1. Are you having issues using `mscs status` with only the `wookies` world? Or does it occur to other worlds as well? Perhaps try creating another world and letting it run for an hour, and then see if the issue also occurs when you try to get the `status` of that world; this could help tell us if it's world-specific or if there's something else going on. The second thing I would like to check is whether or not the amount of users online has something affecting it; did you find that `mscs status` works when, say, there's less than 3 users online and then stops working when there's more than that?  Perhaps the issue centers around how many users are online.

2. Also, are you running MSCS in a multi-user environment / what user(s) are you using to running MSCS? (That is, if you didn't follow the default installation). It could be a permission issue as @sandain noted

Sorry that we can't give you a clear cut answer; right now we have to isolate the problem and then figure it out at this point.

jwilling:
Huge thanks @sandain and @Roflicide for your assistance.

[...]

PR #233, opened 2019-12-15
query server restarts if shut down

sheimer:
query server now restarts on requests, if it shut down before (see issue 232 on https://github.com/MinecraftServerControl/mscs/issues/232)

### DIFF
--- a/msctl
+++ b/msctl
@@ -1753,7 +1753,7 @@ queryStart() {
   # server is piped into the query.out file.  Give the server a moment to
   # start before initializing the Query handler.
   nohup sh -c "
-    sleep 20;
+    sleep 1;
     tail -f --pid=$SERVER_PID \"$WORLD_DIR/query.in\" |
     $SOCAT - UDP:$SERVER_IP:$QUERY_PORT > \"$WORLD_DIR/query.out\"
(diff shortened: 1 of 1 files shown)


conrad784:
this sleep was introduced with #169 , can you debug this behavior with your change?

sheimer:
in #169 are different people having different problems, i guess.

if in the first 20 seconds, while the sleep is sleeping, a "mscs status" is executed, the user receives a message "offline". But since the service is not yet running, this is not a problem, you just wait, and after the sleep all is good.

if with my fix the sleep is removed, and someone asks the status in the first 20 seconds (or rather how ever long it takes the minecraft server to be up and running, and accepting queries),  the socat shuts down. but, on next query it tries to start again, and as soon as minecraft server is up the socat stays in place.

even if socat should stop for some other reasons (maybe minecraft queries not responsive etc.), my fix would just (try) to start it again, so you dont have to restart the whole world, multiple times, like someone explained in #169

sandain:
I'm sorry, I still haven't had time to test this change.  It looks pretty benign so I could probably just accept it, but it would be nice to have some verification that this change works.  Has anyone had the chance to test it yet?

PR #234, opened 2019-12-15
Add new line after each server listed

ScottKillen:
This fixes the output for the 'list' command so that a new line is added after the property listing for each server.


sandain:
Thanks for the patch!

Issue #25, opened 2013-11-29
Netcat -q option fails on CentOS


Firefall-main:
Netcat or nc doesn't support the -q option in CentOS. To prevent errors I had to remove -q 1 from the nc command.

PR #242, opened 2020-05-11
Wait until query port opens before attempting to query

Roflicide:
Hi,

Attempting to query a world before the query port is open results in the socat "connection" (not really a connection since it's UDP) being dropped. Additionally, the code in lines 1734-1736 is never triggered -- even if the query connection failed, the error message will never print. 
```
  # Verify the connection to the query server worked.	
  if [ $? -ne 0 ]; then	
    printf "Error connecting to the query server.\n"	
  fi
````
Subsequent queries to the world will result in a `running (query server offline)` message because the socat connection will never be re-established. 

Isssue #232 fixes this issue downstream by having an `else` clause that executes the `queryStart()` function again. However, this is only triggered when the socat connection has dropped at least once, so it's a bit inefficient.

This pull request fixes the root issue by waiting until the minecraft query port is open through the use of netcat (I assume most linux distros ship with netcat so this shouldn't be a dependency issue, but if there's a way to do the same thing with socat than that may be better).  Once this query port is detected as open, it creates the `query.in` and `query.out` files and establishes the socat connection. By doing it this way, we can detect if a) the world is still starting and that's why we're not getting queries or b) there is something actually wrong with the query server even though the port is open (which is what is implied in Issue #169, though in my testing, this has never happened). 

I'm leaving the code in from #232 (thanks to @sheimer for the PR ! ) as a redundancy in case the query server drops even after the query ports are open (again, as is implied in #169).

### DIFF
--- a/msctl
+++ b/msctl
@@ -1798,7 +1798,7 @@ querySendPacket() {
     '
   # If the response is empty and the query.in and query.out files have NOT been created,
   # the world is still starting up and the status query will return as such. 
-  # See lines 1987-1989.
+  # See lines 1985-1987.
 
   # If the response is empty and the query.in and query.out files HAVE been created, 
   # something went wrong. We try restarting the query handler again.
--- a/msctl
+++ b/msctl
@@ -1737,7 +1737,7 @@ queryStart() {
   # Wait until the query port is open before creating the query files
   # and establishing the tail process.
   nohup sh -c "
-    while ! nc -u -z $SERVER_IP $QUERY_PORT; do
+    while ! echo | $SOCAT - UDP-CONNECT:$SERVER_IP:$QUERY_PORT >/dev/null 2>&1; do
       sleep 0.1
     done;
     printf '' >\"$WORLD_DIR/query.in\"
--- a/msctl
+++ b/msctl
@@ -1731,22 +1731,20 @@ queryStart() {
   SERVER_PID=$(getJavaPID $1)
   # Delete any old query.in or query.out buffer files.
   rm -Rf "$WORLD_DIR/query.in" "$WORLD_DIR/query.out"
-  # Initialize the query.in and query.out buffer files.
-  printf '' >"$WORLD_DIR/query.in"
-  printf '' >"$WORLD_DIR/query.out"
   # Start a tail process to watch for changes to the query.in file to pipe
   # to the Minecraft query server via socat.  The response from the query
-  # server is piped into the query.out file.  Give the server a moment to
-  # start before initializing the Query handler.
+  # server is piped into the query.out file.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by sandain** [comment]
@Roflicide thanks for the patch, but I'm going to give this a NAK. There is a BSD netcat and a GNU netcat that do not have compatible command line arguments. If we are going to switch back to using netcat, we need to incorporate code that detects which version is being run so that the call can be made correctly for both versions.

**Comment by sandain** [comment]
I just read the socat docs, and it looks like your call to netcat could be replaced with something like:
```
- while ! nc -u -z $SERVER_IP $QUERY_PORT; do
+ while ! echo | $SOCAT - UDP-CONNECT:$SERVER_IP:$QUERY_PORT >/dev/null 2>&1; do
```

World on 25667 not running and world on 25668 running:
```
$ echo | socat - UDP-CONNECT:127.0.0.1:25567 >/dev/null 2>&1; echo $?
1
$ echo | socat - UDP-CONNECT:127.0.0.1:25568 >/dev/null 2>&1; echo $?
0
```

Roflicide:
> @Roflicide thanks for the patch, but I'm going to give this a NAK. There is a BSD netcat and a GNU netcat that do not have compatible command line arguments. If we are going to switch back to using netcat, we need to incorporate code that detects which version is being run so that the call can be made correctly for both versions.

I didn't know this was thing, good catch

> I just read the socat docs, and it looks like your call to netcat could be replaced with something like:
> 
> ```
> - while ! nc -u -z $SERVER_IP $QUERY_PORT; do
> + while ! echo | $SOCAT - UDP-CONNECT:$SERVER_IP:$QUERY_PORT >/dev/null 2>&1; do
> ```
> 
> World on 25667 not running and world on 25668 running:
> 
> ```
> $ echo | socat - UDP-CONNECT:127.0.0.1:25567 >/dev/null 2>&1; echo $?
> 1
> $ echo | socat - UDP-CONNECT:127.0.0.1:25568 >/dev/null 2>&1; echo $?
> 0
> ```

nice solution, i tested this and it works! ill add another commit

sandain:
> >@Roflicide thanks for the patch, but I'm going to give this a NAK. There is a BSD netcat and a GNU netcat that do not have compatible command line arguments. If we are going to switch back to using netcat, we need to incorporate code that detects which version is being run so that the call can be made correctly for both versions.
> 
> I didn't know this was thing, good catch

I didn't either until it bit me!  MSCS used to use netcat in the first iteration of the query code

sandain:
See Issue #25 RE the switch to socat

sandain:
I just tried this out and it works nicely.  I'll go ahead and accept the PR.  Thanks for the work on this @Roflicide

PR #245, opened 2020-05-25
Use Perl's switch parsing argument to pass variables to scripts

sandain:
This is in the same line as commit 7c4b4545c07e273a815135fc04c4d1ee3a140f1b.


sandain:
I've been running this patch for a while now, but I'm not sure I have tested everything. Please look over the patch and let me know if you see any issues.

sandain:
More merge errors!  I'm going to start over again.. argh!

Issue #240, opened 2020-03-08
mscs-jvm-args can't contain spaces


zdimension:
If it does, the `start` command fails because the args aren't properly passed to perl, which read something like this:

    -XX:Arg1 -XX:Arg2

as 

    perl -ne '....... -XX:Arg1' -XX:Arg2

giving the error

    Unrecognized switch: -:Arg2    (-h will show valid options).

Problem is in file `msctl`, in `getServerCommand` when inserting the `$JVM_ARGS` variable in the perl call

sandain:
Sorry for the slow response.  Commit 7c4b4545c07e273a815135fc04c4d1ee3a140f1b should fix this issue.

PR #246, opened 2020-05-26
Use Perl's switch parsing argument to pass variables to scripts

sandain:
This is in the same line as commit 7c4b4545c07e273a815135fc04c4d1ee3a140f1b.

### DIFF
--- a/msctl
+++ b/msctl
@@ -318,8 +318,9 @@ getJavaPID() {
   local PID MCUSER
   MCUSER=$(whoami)
   PID=$(
-    ps -a -U $MCUSER -o pid,comm,args -ww |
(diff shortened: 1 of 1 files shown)


sandain:
I've been running this patch for a while now, but I'm not sure I have tested everything. Please look over the patch and let me know if you see any issues.

sandain:
This is a more robust way of passing variables into Perl. I made this patch for two reasons, i) to use the same pattern as was done in commit 7c4b4545c07e273a815135fc04c4d1ee3a140f1b, that fixed an issue (#240) with parsing spaces in properties files, to simplify future code debugging, and ii) it actually fixes a bug I found while working on new code for msc-gui.

Roflicide:
i see, very nice !

PR #248, opened 2020-06-10
Add notice when attempting to start/stop an already started/stopped world

Roflicide:
Attempting to start an already started world or stop an already stopped world will only display a period ("."), such as below:

````
$ mscs status
Minecraft Server Status:
  test: running version 1.15.2 (2 of 20 users online).
    Process ID: 5160.
    Memory used: 2538068 kB.
  cannon: not running.

$ mscs start test
The cached copy of the version manifest is up to date.
Use the force-update option to ensure a new copy is downloaded.
Starting Minecraft Server:.

$ mscs stop cannon
Stopping Minecraft Server:.

````

This patch notifies the user when they attempt to start or stop an already started/stopped world. If the user tries to start/stop multiple worlds and one already is started/stopped, then the entire command will exit and no worlds will be started or stopped (same behavior when user tries to start/stop/restart worlds and one world doesn't exist).

````
$ mscs status
Minecraft Server Status:
  test: running version 1.15.2 (2 of 20 users online).
    Process ID: 5160.
    Memory used: 2538068 kB.
  cannon: not running.

$ mscs start test
The cached copy of the version manifest is up to date.
Use the force-update option to ensure a new copy is downloaded.
Unable to start the requested worlds: world 'test' already running.

$ mscs stop cannon
Unable to stop the requested worlds: world 'cannon' already stopped.

$ mscs start test cannon
The cached copy of the version manifest is up to date.
Use the force-update option to ensure a new copy is downloaded.
Unable to start the requested worlds: world 'test' already running.

$ mscs stop test cannon
Unable to stop the requested worlds: world 'cannon' already stopped.
````

Additionally, right now when a user tries to start/stop/restart a world that doesn't exist, it will say "World not recognized . I changed the warnings to "World not recognized **or disabled**" since disabled worlds won't be recognized either (since the worlds are being checked with the `isWorldEnabled` function).

### DIFF
--- a/msctl
+++ b/msctl
@@ -2416,13 +2416,13 @@ case "$COMMAND" in
       for arg in "$@"; do
         if isWorldEnabled "$arg"; then
           if serverRunning "$arg"; then
-            printf "Unable to start the requested world(s): world '$arg' already running.\n"
+            printf "Unable to start the requested worlds: world '$arg' already running.\n"
             exit 1
           else
             WORLDS="$WORLDS $arg"
           fi
         else
-          printf "World '$arg' not recognized or disabled.\n"
+          printf "Unable to start the requested worlds: world '$arg' not recognized or disabled.\n"
           printf "  Usage:  $PROG $COMMAND <world1> <world2> <...>\n"
           exit 1
         fi
@@ -2444,13 +2444,13 @@ case "$COMMAND" in
       for arg in "$@"; do
         if isWorldEnabled "$arg"; then
           if ! serverRunning "$arg"; then
-            printf "Unable to stop the requested world(s): world '$arg' already stopped.\n"
+            printf "Unable to stop the requested worlds: world '$arg' already stopped.\n"
             exit 1
           else
             WORLDS="$WORLDS $arg"
           fi
         else
-          printf "World '$arg' not recognized or disabled.\n"
+          printf "Unable to stop the requested worlds: world '$arg' not recognized or disabled.\n"
           printf "  Usage:  $PROG $COMMAND <world1> <world2> <...>\n"
           exit 1
         fi
@@ -2497,7 +2497,7 @@ case "$COMMAND" in
         if isWorldEnabled "$arg"; then
           WORLDS="$WORLDS $arg"
         else
-          printf "World '$arg' not recognized or disabled.\n"
+          printf "Unable to restart the requested worlds: world '$arg' not recognized or disabled.\n"
           printf "  Usage:  $PROG $COMMAND <world1> <world2> <...>\n"
           exit 1
         fi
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by zanix** [comment]
Having the `force-*` variants skip this check might be a good idea.

**Comment by Roflicide** [comment]
> Having the `force-*` variants skip this check might be a good idea.

oh, good point. i totally forgot about those commands

**Comment by zanix** [comment]
I don't think there is a force-start though

File: msctl [hunk 1]
```diff
@@ -2487,7 +2497,7 @@ case "$COMMAND" in
         if isWorldEnabled "$arg"; then
           WORLDS="$WORLDS $arg"
         else
```
sandain, line 2499: Maybe we should add in an `elif isWorldDisabled "$arg"; then` section..

sandain:
I like this, good idea.

msctl
Roflicide: > Maybe we should add in an `elif isWorldDisabled "$arg"; then` section..

that could be a good addition. do we wanna do it for just the start/stop/restart commands or for all of the commands?

sandain: I was just thinking about what was already being touched here. Instead of adding "or disabled", I thought it would be nicer to report separate error messages.

Roflicide: ok, i just added commit a0c0f6d that adds the notifications for trying to start/stop/restart disabled worlds

Roflicide: do we want the `force-stop` and `force-restart` commands to skip this check like @zanix eluded to?

sandain:
Nice. Thanks @Roflicide. I'll accept the PR.

Roflicide:
Woops -- I accidently had in code there that i was gonna put in another PR that warns the user before checking to delete a world (bold is what I added to the delete command)

do you want me revert this back?
````
delete | remove)
    # Get list of enabled worlds.
    if ! isWorldAvailable "$1"; then
      printf "World not found, unable to delete world '$1'.\n"
      exit 1
    fi
    **printf "Are you sure you wish to delete world '$1'? (y/n): "
    read answer
    if [ "$answer" = "${answer#[Yy]}" ]; then
        printf "Aborted.\n"
        exit 1
    else**
      printf "Deleting Minecraft world: $1"
      if serverRunning "$1"; then
        # If the world server has users logged in, announce that the world is
        # being deleted.
        if [ $(queryNumUsers "$1") -gt 0 ]; then
          sendCommand "$1" "say The server admin is deleting this world."
          sendCommand "$1" "say The server will be deleted in 1 minute..."
          sleep 60
          sendCommand "$1" "say The server is now shutting down."
        fi
        # Stop the world server.
        stop "$1"
        sleep 5
      fi
      # Delete the world.
      deleteWorld "$1"
      printf ".\n"
    fi
````

sandain:
That is a good idea. Don't worry about it being in this PR

Issue #225, opened 2019-09-30
adding a way to monitor server crash and automatically restart


nirim000:
I try to write a start script for spigot to restart through mscs in case of a crash,
 no matter what i do spigot is unable to start the script.

I'm currently relying on a cronjob to try to start the server in case of a crash.

Adding the ability to monitor crash and automatically restart after a crash to mscs would by most helpful. 

I found a interactive script that seem like a good base https://gist.github.com/esolitos/af3e045552a87b513715e854d4a3fe16

sorry for my terrible formatting, i'm horribly tired.

Roflicide:
@sandain I'm working on this enhancement but I've ran into an issue:

My initial approach to tackle this problem was to spin off a child shell via `nohup` at the end the `start()` function that watches the server and restarts it if it detects that it crash. The problem is that only `bash` allows you to export functions to child shells; `sh` doesn't (and the shebang ontop of mscs is `sh`). So, I would have to copy/paste of all the functions and dependencies of the `start` command into the child shell code, which would make it quite lengthy and redundant. Any ideas on how to get around this?

sandain:
Could the sub process just call `mscs start` or whatever? Why do you need to export functions?

Roflicide:
oh yes, your right that would work for start...im not sure why i didnt think of that. good catch. the only issue is that some of the functions i call don't have an associated command, but theres only a few of them. ill submit the PR soon so maybe you can get a better idea of what im talking about

PR #257, opened 2020-08-16
Add automatic crash detection and restarts

Roflicide:
## Automatic crash detection and restarts

```
; Enable the option to restart the server after a crash is detected.
; 0 - Restart after crash (default).
; 1 - Do not restart after crash.
# mscs-enable-restart-after-crash=0
```

This is a preliminary PR for #225 that enables the option to restart worlds after a crash is detected using the `mscs-enable-restart-after-crash` flag (via `mscs.defaults`). I set the default to be 0 (`true`), but we can change the default to be 1 (`false`) if you think that's better. Additionally, we could add the functionality for it to be available on a per world basis via `mscs.properties` if you think that's a good idea as well.

How it works:

1. When a world is started, the `startServerMonitor()` function is called. This checks if the `mscs-enable-restart-after-crash` flag is enabled. If so, it checks if there is another instance of the server monitor already running (which would occur if the server was restarted after a crash). To do this check, it uses `flock`. I tried for several hours to use the `createLockFile()` function, but it didn't work for me for whatever reason--not sure if it was because I was doing something wrong or if it was just an issue with the function itself. @sandain  If you're able to get the `createLockFile()` function working instead of using flock, that'd be good so we don't have to use `flock` as a dependency. 

2. If the monitor is not running, the `startServerMonitor()` function essentially calls the `nohup` command on the `serverMonitor()` function and runs it in the background, so you can close the terminal and the crash monitor will keep running. `nohup` doesn't allow you to pass functions, but I managed to recreate the functionality through some research. The PID of the monitor is stored in `$WORLD_DIR/monitor.pid`.

3. The `serverMonitor()` function is essentially a loop that detects if the server has crashed by a combination of the a) the existence of the `$WORLD.pid` file and b) whether the server is running via the `serverRunning()` function. If the PID file exists but the server isn't running, then this means the server probably crashed (since MSCS hasn't removed the PID file). 
Additionally, I had to modify the serverRunning() function to use `echo` instead of `return` to return values so I could call the function in the test condition of the `until` loop and check the return value using `$()` (I didn't think I could do this the way the function was, but maybe I'm wrong). I subsequently modified all of the references in the script to call serverRunning using `$()`.
 There is also a log for the monitor written to `$WORLD_DIR/logs/mscs.monitor.log`. This log is deleted after `$LOG_DURATION` to make sure it doesn't get too large. This check occurs in the `startServerMonitor()` function. There is also a log created called `last-start-status.log` that stores a message to be displayed later in the `mscs status` command.

4. When the `mscs stop` command is called, it sends a request to kill the server monitor using the `stopServerMonitor()` function. At this point the monitor should be killed.

When a user runs the `mscs status` for the first time after a server has crashed, they will see a notice that the world has crashed:
```
$ mscs status
Minecraft Server Status:
  alex: running version 1.16.2 (0 of 20 users online).
    Port: 25567.
    Memory used: 1.04GB (2GB allocated).
    Server PID: 3351.
    Crash Monitor PID: 19569
    alex automatically restarted from a crash (or in-game stop command)
    on 2020-08-15_22-49-13. See
    /opt/mscs/worlds/alex/logs/mscs.monitor.log and
    /opt/mscs/worlds/alex/crash_reports/ and
    /opt/mscs/worlds/alex/logs/ for more information.
```
This notice goes away after the first time. The code behind this is simply a check if whether or not the `last-start-status.log` file exists and is not empty. If this is true, the message is displayed and the file is deleted afterwards (so the message is only displayed once).

P.S.: I tested crashes on the code by doing the `/stop` command in-game, I believe it should replicate what happens with an actual crash.

Please let me know of any questions or recommendations you have or bugs you find. Thanks!!

**Summary of TODO:**
- Remove flock dependency
- Test more
- Question: add functionality to enable it on a per-world basis?
- Question: make the default setting true or false?

### DIFF
--- a/msctl
+++ b/msctl
@@ -2323,7 +2323,7 @@ fi
 #   mscs-default-maximum-memory     - Default maximum amount of memory for a world server.
 #   mscs-default-server-location    - Default location of the server .jar file.
 #   mscs-default-server-command     - Default command to run for a world server.
-#   mscs-enable-restart-after-crash - Enable the option to restart worlds after crash detected (default enabled).
+#   mscs-enable-restart-after-crash - Enable the option to restart the server after crash detected (default enabled).
 #   mscs-backup-location            - Location to store backup files.
 #   mscs-backup-log                 - Location of the backup log file.
 #   mscs-backup-duration            - Length in days that backups survive.
--- a/msctl
+++ b/msctl
@@ -1331,9 +1331,11 @@ serverMonitor() {
       # Verify that the server restarted successfully. 
       if [ $? -eq 0 ]; then
         printf "[$(timestamp)] [INFO]: Server monitoring resumed for $1. Server PID: $(getJavaPID $1). Monitor PID: $MONITOR_PID.\n"
-        printf "    $1 automatically restarted from a crash on $(timestamp).\n" > "$LAST_START_STATUS_LOG"
-        printf "    See $WORLD_DIR/logs/mscs.monitor.log and\n" >> "$LAST_START_STATUS_LOG"
-        printf "    $WORLD_DIR/crash_reports/ for more information.\n" >> "$LAST_START_STATUS_LOG"
+        printf "    $1 automatically restarted from a crash (or in-game stop command)\n" > "$LAST_START_STATUS_LOG"
+        printf "    on $(timestamp). See\n" >> "$LAST_START_STATUS_LOG"
+        printf "    $WORLD_DIR/logs/mscs.monitor.log and\n" >> "$LAST_START_STATUS_LOG"
+        printf "    $WORLD_DIR/crash_reports/ and\n" >> "$LAST_START_STATUS_LOG"
+        printf "    $WORLD_DIR/logs/ for more information.\n" >> "$LAST_START_STATUS_LOG"
       else
         printf "[$(timestamp)] [ERROR]: Failed to restart $1.\n" 
         stopServerMonitor $1
@@ -2323,7 +2325,7 @@ fi
 #   mscs-default-maximum-memory     - Default maximum amount of memory for a world server.
 #   mscs-default-server-location    - Default location of the server .jar file.
 #   mscs-default-server-command     - Default command to run for a world server.
-#   mscs-enable-restart-after-crash - Enable the option to restart the server after crash detected (default enabled).
+#   mscs-enable-restart-after-crash - Enable the option to restart server after crash detected (default enabled).
 #   mscs-backup-location            - Location to store backup files.
 #   mscs-backup-log                 - Location of the backup log file.
 #   mscs-backup-duration            - Length in days that backups survive.
--- a/msctl
+++ b/msctl
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # ---------------------------------------------------------------------------
-# Copyright (c) 2011-2016, Jason M. Wood <sandain@hotmail.com>
+# Copyright (c) 2011-2020, Jason M. Wood <sandain@hotmail.com>
 #
 # All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


zanix:
`0` should be disabled and `1` should be enabled

sandain:
I agree with @zanix, 0 should be disabled and 1 enabled. It would be confusing otherwise. A per world option in addition to the global option would be best. Default should be disabled I think.

Sorry about the slow review. I've been busy.

msctl
```diff
@@ -346,9 +351,9 @@ getJavaMemory() {
 serverRunning() {
   # Try to determine if the world is running.
   if [ $(getJavaPID "$1") -gt 0 ]; then
-    return 0
+    echo 0
   else
-    return 1
+    echo 1
```
sandain, line 356: I don't think this change is necessary. It forces a lot of other changes that make the code less readable.

E.g.
```
-  while serverRunning $1; do
+  while [ $(serverRunning $1) -eq 0 ]; do
```

sandain (changes requested):

msctl
```diff
@@ -1262,6 +1267,122 @@ serverConsole() {
...
+  WORLD_DIR="$WORLDS_LOCATION/$1"
+  MONITOR_LOG="$WORLD_DIR/logs/mscs.monitor.log"
+  SERVER_LOG="$WORLD_DIR/logs/latest.log"
+  LAST_START_STATUS_LOG="$WORLD_DIR/logs/last-start-status.log"
+  MONITOR_PID=$(cat "$WORLD_DIR/monitor.pid")
+  touch $LAST_START_STATUS_LOG
+
+  printf "[$(timestamp)] [INFO]: Server monitoring started for $1. Server PID: $(getJavaPID $1). Monitor PID: $MONITOR_PID.\n"
+  # Run monitor until the server is stopped and the PID file is removed (i.e. clean shutdown).
+  until [ $(serverRunning $1) -eq 1 ] && [ ! -f "$WORLDS_LOCATION/$1.pid" ]; do
```
sandain, line 1326: This line would become:
```
until [ ! serverRunning $1 ] && [ ! -f "$WORLDS_LOCATION/$1.pid" ]; do
```

Which I think is more readable.

msctl
sandain: Usually 0 means false, and 1 means true. UNIX land returns 0 when a command successfully ran and a value >0 for error states. This return trick was used here to make the code more readable, but the values are inverted from what you mean when changed to echo.

Roflicide: Thanks for clarifying this. It's confusing because, as you said, in UNIX land 0 is returned when the command ran successfully. In my head, I associate success with being "TRUE", so it makes me think it should return 1 based on the 0=false and 1=true convention, but it doesn't.... But anyway, I think I understand now, thanks

sandain: We should probably add comments to the serverRunning method to note this UNIX oddity.

```patch
serverRunning() {
  # Try to determine if the world is running.
  if [ $(getJavaPID "$1") -gt 0 ]; then
-    return 0
+    return 0  # Return a true value.
  else
-    return 1
+    return 1  # Return a false value.
  fi
}

```

msctl
Roflicide: yes, agreed

msctl
Roflicide: @sandain do you know why `if [serverRunning $1]` doesn't work? it only works when i omit the brackets and I can't figure out how to do it if I want to negate the expression

sandain: You need spaces next to the square brackets.
```
if [ serverRunning $1 ]
```

to negate:
```
if [ ! serverRunning $1 ]
```

sandain: Although, I think the brackets are optional.

```
if ! serverRunning $1; then
```

Roflicide:
I'm going to close this PR and re-open one that is up-to-date with the current code.

PR #260, opened 2020-09-15
Added a config option for excluding files from being backed up

broskisworld:
This is a relatively small change that adds an additional config option `mscs-backup-excluded-files` in mscs.defaults (optionally mscs.properties) for excluding files and directories from being backed up.

I'd imagine the primary use for this would be for games with map plugins, though it could also be used for anything in the world folder. Even with differential backups, servers with overviewer or dynmap can make for extremely large backups, despite the fact the fact that tilemap renders can be re-generated from the world file itself in the event of a server crash.

This additional config allows admins higher control over exactly how much space they want to use for backups and enables admins with smaller hard drives to still keep limited backups of critical user data.

The property accepts an individual item or a comma-separated list, and evaluates`$WORLD_NAME` to the name of the world being backed up. I've added a few example use cases below:
 - `mscs-backup-excluded-files=plugins/dynmap/web/tiles` excludes dynmap tiles only
 - `mscs-backup-excluded-files=plugins/dynmap,$WORLD_NAME_nether,$WORLD_NAME_the_end` excludes the entire dynmap configuration, as well as the nether and the end world files
 - `mscs-backup-excluded-files=` default configuration (keep everything)

### DIFF
--- a/msctl
+++ b/msctl
@@ -1446,12 +1446,19 @@ worldBackup() {
   "
   # Add all user-specified excluded files
   OLD_IFS=$IFS
+  IFS=,
   # Check for world-specific excluded files
   WORLD_BACKUP_EXCLUDED_FILES=$(getMSCSValue $1 "mscs-backup-excluded-files" "$BACKUP_EXCLUDED_FILES")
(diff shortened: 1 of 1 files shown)


sandain:
This PR looks good, thanks for the patches. I see how this could be useful.

PR #262, opened 2020-10-02
Suggest Some improvement in CODE_OF_CONDUCT.md

faisalpucit:


sandain:
Thank you for the patch, but I am going to NACK this to keep our code of conduct in line with the upstream version.  If you would like to submit a PR to update the code of conduct to the upstream version 2.0, I would consider it.

PR #266, opened 2020-11-25
Update MSCS Default Properties File and Comments

zanix:
- Add missing `$JVM_ARGS` in `mscs-default-server-command` default property
- Add info for properties that set replacement variables in `mscs_defaults()`
- Moved property keys in global list to match per-world list
- Add missing `$JVM_ARGS` in variables comment
- Add missing `$JVM_ARGS` variable to `mscs-default-server-command` and `mscs-server-command` property keys

### DIFF
--- a/msctl
+++ b/msctl
@@ -217,8 +217,9 @@ mscs_defaults() {
 ; Default version type (release or snapshot).
 # mscs-default-version-type=release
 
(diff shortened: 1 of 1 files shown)


sandain:
This looks good. Thanks for the work

PR #271, opened 2020-12-26
Add automatic crash detection and restarts

Roflicide:
Hi,

This is an updated PR from #257 : New changes:

- Updated the code to reflect the most recent codebase
- Changed the default setting to now being disabled
- Changed the name of the flag (from `mscs-enable-restart-after-crash` to `mscs-restart-after-crash`) for brevity
- Added the ability to set it on a per world basis (`mscs-default-restart-after-crash` and `mscs-restart-after-crash`)
- Changed the flags from `0/1` to `false/true` for clarity's sake
- Removed the old changes I made (in the `if` statements when trying to test if the server was running) that made the code less readable, and reverted it back to how the old way was done
- Updated the copyright year ;)

I still need to figure out if there's a way to check if there's a server monitor process running without using flock (which the code uses right now). As I mentioned in the last PR, I tried using the `createLockFile()` function that's in the script but it didn't work for me for some reason.

Probably should also test it a bit more. It does seem to be working as intended on my computer, though!

Thanks

### DIFF
--- a/msctl
+++ b/msctl
@@ -1323,7 +1323,7 @@ stopServerMonitor() {
 # @param 1 The world server to monitor.
 # ---------------------------------------------------------------------------
 serverMonitor() {
-  local WORLD_DIR MONITOR_LOG SERVER_LOG LAST_START_STATUS_LOG MONITOR_PID GREP_STOP
+  local WORLD_DIR MONITOR_LOG SERVER_LOG LAST_START_STATUS_LOG MONITOR_PID
   WORLD_DIR="$WORLDS_LOCATION/$1"
   MONITOR_LOG="$WORLD_DIR/logs/mscs.monitor.log"
   SERVER_LOG="$WORLD_DIR/logs/latest.log"
--- a/msctl
+++ b/msctl
@@ -47,6 +47,7 @@ WGET=$(which wget)
 RDIFF_BACKUP=$(which rdiff-backup)
 RSYNC=$(which rsync)
 SOCAT=$(which socat)
+FLOCK=$(which flock)
 
 # Script Usage
 # ---------------------------------------------------------------------------
@@ -1310,7 +1311,7 @@ stopServerMonitor() {
       printf "[$(timestamp)] [ERROR]: Unable to kill monitor process.\n" >> "$MONITOR_LOG"
       exit 1
     else 
-      printf "[$(timestamp)] [INFO]: Server monitor process killed sucessfully.\n" >> "$MONITOR_LOG"
+      printf "[$(timestamp)] [INFO]: Server monitor process killed successfully.\n" >> "$MONITOR_LOG"
       # Remove the monitor PID file. 
       rm -f "$WORLD_DIR/monitor.pid"
     fi
@@ -2198,7 +2199,7 @@ worldStatusJSON() {
 # ---------------------------------------------------------------------------
 
 # Make sure that Java, Perl, libjson-perl, libwww-perl, Python, Wget,
-# Rdiff-backup, Rsync, and Socat are installed.
+# Rdiff-backup, Rsync, Socat and flock are installed.
 # ---------------------------------------------------------------------------
 if [ ! -e "$JAVA" ]; then
   echo "ERROR: Java not found!"
@@ -2261,6 +2262,12 @@ if [ ! -e "$SOCAT" ]; then
   echo "ERROR: socat not found!"
   echo "Try installing this with:"
   echo "sudo apt-get install socat"
+  exit 1
+fi
+if [ ! -e "$FLOCK" ]; then
+  echo "ERROR: flock not found!"
+  echo "Try installing this with:"
+  echo "sudo apt-get install flock"
   exit 1
 fi
 
--- a/msctl
+++ b/msctl
@@ -2267,7 +2267,7 @@ fi
 if [ ! -e "$FLOCK" ]; then
   echo "ERROR: flock not found!"
   echo "Try installing this with:"
-  echo "sudo apt-get install flock"
+  echo "sudo apt-get install util-linux"
   exit 1
 fi
 
--- a/msctl
+++ b/msctl
@@ -1300,7 +1300,7 @@ stopServerMonitor() {
   MONITOR_LOCK_FILE="$WORLD_DIR/monitor.lock"
   # Check if server monitor instance currently running.
   (
-  flock -n 9
+  $FLOCK -n 9
   ACQUIRED_LOCK=$?
   if [ "$ACQUIRED_LOCK" -eq 1 ]; then # Server monitor is running.
     printf "[$(timestamp)] [INFO]: Stop command received for server monitor. Attempting to kill server monitor...\n" >> "$MONITOR_LOG"
--- a/msctl
+++ b/msctl
@@ -1376,7 +1376,7 @@ startServerMonitor() {
   if [ "$RESTART_AFTER_CRASH" = "true" ]; then
     # Verify that there is no monitor instance currently running.
     (
-    flock -n 9
+    $FLOCK -n 9
     ACQUIRED_LOCK=$?
     if [ "$ACQUIRED_LOCK" -eq 0 ]; then # Server monitor doesn't exist.
       # Delete old log file greater than $LOG_DURATION, if it exists.
--- a/msctl
+++ b/msctl
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # ---------------------------------------------------------------------------
-# Copyright (c) 2011-2016, Jason M. Wood <sandain@hotmail.com>
+# Copyright (c) 2011-2021, Jason M. Wood <sandain@hotmail.com>
 #
 # All rights reserved.
 #
@@ -275,6 +275,9 @@ mscs_defaults() {
 ; selected. The \$SERVER_ARGS variable provides access to the server arguments
 ; for the world server selected.
 # mscs-default-server-command=\$JAVA -Xms\$INITIAL_MEMORY -Xmx\$MAXIMUM_MEMORY \$JVM_ARGS -jar \$SERVER_LOCATION/\$SERVER_JAR \$SERVER_ARGS
+
+; Default behavior if to restart the server after crash is detected (default disabled).
+# mscs-default-restart-after-crash=false
 
 ; Location to store backup files.
 # mscs-backup-location=/opt/mscs/backups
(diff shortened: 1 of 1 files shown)


sandain:
Thanks for the new PR @Roflicide.  This looks much better!

I had to rebuild my server a month ago due to a failed drive, and haven't yet installed MSCS (I haven't even played Minecraft in what seems like years). I'll try to make some time soon to give this a try.  If @zanix or anyone else has the chance to test this out, and approves it before I get the chance, feel free to merge the PR.

msctl
```diff
@@ -2186,46 +2326,50 @@ fi
...
+#   mscs-default-server-version      - Default version of the server software.
+#   mscs-default-jvm-args            - Default arguments for the JVM.
+#   mscs-default-server-jar          - Default .jar file for the server software.
+#   mscs-default-server-url          - Default download URL for the server software.
+#   mscs-default-server-args         - Default arguments for a world server.
+#   mscs-default-initial-memory      - Default initial amount of memory for a world server.
+#   mscs-default-maximum-memory      - Default maximum amount of memory for a world server.
+#   mscs-default-server-location     - Default location of the server .jar file.
+#   mscs-default-server-command      - Default command to run for a world server.
+#   mscs-default-server-command      - Default command to run for a world server.
```
zanix, line 2359: This is a duplicate of the line above

msctl
```diff
@@ -2186,46 +2326,50 @@ fi
...
-#   mscs-default-initial-memory  - Default initial amount of memory for a world server.
-#   mscs-default-maximum-memory  - Default maximum amount of memory for a world server.
-#   mscs-default-server-location - Default location of the server .jar file.
-#   mscs-default-server-command  - Default command to run for a world server.
-#   mscs-backup-location         - Location to store backup files.
-#   mscs-backup-log              - Lcation of the backup log file.
-#   mscs-backup-excluded-files   - Comma separated list of files and directories excluded from backups.
-#   mscs-backup-duration         - Length in days that backups survive.
-#   mscs-log-duration            - Length in days that logs survive.
-#   mscs-detailed-listing        - Properties to return for detailed listings.
```
zanix, line 2222: This was also moved here in an earlier PR so it matches the default value listing section below

msctl
```diff
@@ -1274,6 +1277,123 @@ serverConsole() {
...
+startServerMonitor() {
+  local WORLD_DIR MONITOR_LOG MONITOR_PID MONITOR_LOCK_FILE ACQUIRED_LOCK
+  WORLD_DIR="$WORLDS_LOCATION/$1"
+  MONITOR_LOG="$WORLD_DIR/logs/mscs.monitor.log"
+  MONITOR_PID="$WORLD_DIR/monitor.pid"
+  MONITOR_LOCK_FILE="$WORLD_DIR/monitor.lock"
+  RESTART_AFTER_CRASH=$(getMSCSValue "$1" "mscs-restart-after-crash" "$DEFAULT_RESTART_AFTER_CRASH")
+
+  # Verify option is enabled.
+  if [ "$RESTART_AFTER_CRASH" = "true" ]; then
```
zanix, line 1375: If someone uses `1` for the `mscs-restart-after-crash/mscs-default-restart-after-crash` does this statement still evaluate as `true`

msctl
```diff
@@ -276,6 +276,9 @@ mscs_defaults() {
 ; for the world server selected.
 # mscs-default-server-command=\$JAVA -Xms\$INITIAL_MEMORY -Xmx\$MAXIMUM_MEMORY \$JVM_ARGS -jar \$SERVER_LOCATION/\$SERVER_JAR \$SERVER_ARGS
 
+; Default behavior if to restart the server after crash is detected (default disabled).
+# mscs-default-restart-after-crash=false
```
zanix, line 280: `true/false` is not used for any other mscs setting, will using `0/1` still evaluate to `true/false`?

msctl
```diff
@@ -1274,6 +1277,123 @@ serverConsole() {
...
+# ---------------------------------------------------------------------------
+stopServerMonitor() {
+  local WORLD_DIR MONITOR_LOG MONITOR_PID MONITOR_LOCK_FILE ACQUIRED_LOCK
+  WORLD_DIR="$WORLDS_LOCATION/$1"
+  MONITOR_LOG="$WORLD_DIR/logs/mscs.monitor.log"
+  MONITOR_PID="$WORLD_DIR/monitor.pid"
+  MONITOR_LOCK_FILE="$WORLD_DIR/monitor.lock"
+  # Check if server monitor instance currently running.
+  (
+  flock -n 9
```
zanix, line 1302: `flock` should to be added to the `# Required Software` checks section and should use a variable alias

zanix (changes requested):
Can you also make a PR for the documentation to add the new mscs setting options?

msctl
```diff
@@ -2186,46 +2326,50 @@ fi
...
-#   mscs-backup-duration         - Length in days that backups survive.
-#   mscs-log-duration            - Length in days that logs survive.
-#   mscs-detailed-listing        - Properties to return for detailed listings.
-#   mscs-enable-mirror           - Enable the mirror option by default for worlds (default disabled).
-#   mscs-mirror-path             - Default path for the mirror files.
-#   mscs-overviewer-bin          - Location of Overviewer.
-#   mscs-overviewer-url          - URL for Overviewer.
-#   mscs-maps-location           - Location of Overviewer generated map files.
-#   mscs-maps-url                - URL for accessing Overviewer generated maps.
+#   mscs-location                    - Location of the mscs files.
```
zanix, line 2335: This whole section seems to have been changed due to whitespaces changes

msctl
sandain: Good point. We probabably should add code like:

```bash
#!/bin/sh

true_value() {
  local VALUE
  VALUE=$(echo "$1" | tr '[:upper:]' '[:lower:]')
  if [ "$VALUE" = "true" ] || [ $VALUE = 1 ]; then
    return 0
  fi
  if [ "$VALUE" = "false" ] || [ $VALUE = 0 ]; then
    return 1
  fi
  return -1
}

if true_value "true"; then echo "true"; else echo "false?"; fi     # Prints true
if true_value "TRUE"; then echo "true"; else echo "false?"; fi     # Prints true
if true_value 1; then echo "true"; else echo "false?"; fi          # Prints true
if true_value "1"; then echo "true"; else echo "false?"; fi        # Prints true


if ! true_value "false"; then echo "false"; else echo "true?"; fi  # Prints false
if ! true_value "FALSE"; then echo "false"; else echo "true?"; fi  # Prints false
if ! true_value 0; then echo "false"; else echo "true?"; fi        # Prints false
if ! true_value "0"; then echo "false"; else echo "true?"; fi      # Prints false

if true_value "cat"; then echo "cat?"; else echo "not cat?";  fi   # Prints error message
```

zanix: Very nice, might be easier for users to be able to use `true/false` over `1/0`

zanix: Do you think expanding this to accept `on/off` would be useful?

zanix: Or heck, even `yes/no`

I better stop before going off the rails...

sandain: That is doable...

```bash
true_value() {
  local VALUE
  VALUE=$(echo "$1" | tr '[:upper:]' '[:lower:]')
  if [ $VALUE = 1 ] || [ "$VALUE" = "true" ] || [ "$VALUE" = "yes" ] || [ "$VALUE" = "on" ]; then
    # Return a true value (success).
    return 0
  fi
  if [ $VALUE = 0 ] || [ "$VALUE" = "false" ] || [ "$VALUE" = "no" ] || [ "$VALUE" = "off" ]; then
    # Return a false value (failure).
    return 1
  fi
  return -1
}
```

sandain: I just created PR #274 to add this `true_value` code

msctl
sandain: There is a method in the script already called `createLockFile`. Would this work instead of using `flock`?

Roflicide: Initially tried to use `createLockFile` instead of `flock` but it didn't work. If you are able to get it working without `flock` that would be great

msctl
Roflicide: yes, I increased the white space because the new variable names are long and didn't fit nicely into the column before

msctl
Roflicide: thanks for the catch

msctl
Roflicide: got it, fixed. thank you

msctl
```diff
@@ -2186,46 +2326,50 @@ fi
...
-#   mscs-default-jvm-args        - Default arguments for the JVM.
-#   mscs-default-server-jar      - Default .jar file for the server software.
-#   mscs-default-server-url      - Default download URL for the server software.
-#   mscs-default-server-args     - Default arguments for a world server.
-#   mscs-default-initial-memory  - Default initial amount of memory for a world server.
-#   mscs-default-maximum-memory  - Default maximum amount of memory for a world server.
-#   mscs-default-server-location - Default location of the server .jar file.
-#   mscs-default-server-command  - Default command to run for a world server.
-#   mscs-backup-location         - Location to store backup files.
-#   mscs-backup-log              - Lcation of the backup log file.
```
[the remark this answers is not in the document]
Roflicide: got it, fixed. thank you

msctl
Roflicide: thanks for this!

msctl
Roflicide: i believe @sandain PR #274 addresses this

Roflicide (commented):
hi @zanix, thanks for leaving some feedback. I believe I've fixed all of your requested edits with the exception of the added white space, as I think it's necessary it since the `mscs-default-restart-after-crash` variable is so long. let me know if you find anything else

thanks!

msctl
zanix: It does indeed

msctl
zanix: Ah, I didn't catch that.

msctl
zanix: What didn't work when using `createLockFile`?

You are using `-n` which immediately exits with `1` and doesn't wait for a lock, so I think you could rewrite this to use `createLockFile` however, this function only creates a pid file

zanix: `createLockFile` returns false when a lock exists, you might need to reverse the logic

zanix: But it could return false if the lock file was not created...

msctl
```diff
@@ -1274,6 +1277,123 @@ serverConsole() {
...
+  if [ "$ACQUIRED_LOCK" -eq 1 ]; then # Server monitor is running.
+    printf "[$(timestamp)] [INFO]: Stop command received for server monitor. Attempting to kill server monitor...\n" >> "$MONITOR_LOG"
+    # Kill the server monitor.
+    kill -9 $(cat $MONITOR_PID)
+    # Verify it was actually killed.
+    if [ $? -eq 1 ]; then
+      printf "[$(timestamp)] [ERROR]: Unable to kill monitor process.\n" >> "$MONITOR_LOG"
+      exit 1
+    else 
+      printf "[$(timestamp)] [INFO]: Server monitor process killed sucessfully.\n" >> "$MONITOR_LOG"
```
zanix, line 1313: sucessfully => successfully

msctl
```diff
@@ -1274,6 +1277,123 @@ serverConsole() {
...
+      # Remove the monitor PID file. 
+      rm -f "$WORLD_DIR/monitor.pid"
+    fi
+  fi
+  ) 9>"$MONITOR_LOCK_FILE"
+}
```
zanix, line 1319: This was kinda quick, but does this work?
```suggestion
  WORLD=$1
  MONITOR_LOCK_FILE=$WORLD_DIR/lock-monitor.pid
  # Check if server monitor instance currently running.
  if [ ! $(createLockFile $WORLD "monitor" ) ]  && [ -f "$MONITOR_LOCK_FILE" ]; then # Server monitor is running.
    printf "[$(timestamp)] [INFO]: Stop command received for server monitor. Attempting to kill server monitor...\n" >> "$MONITOR_LOG"
    # Kill the server monitor.
    kill -9 $(cat $MONITOR_LOCK_FILE)
    # Verify it was actually killed.
    if [ $? -eq 1 ]; then
      printf "[$(timestamp)] [ERROR]: Unable to kill monitor process.\n" >> "$MONITOR_LOG"
      exit 1
    else 
      printf "[$(timestamp)] [INFO]: Server monitor process killed successfully.\n" >> "$MONITOR_LOG"
      # Remove the monitor PID file. 
      rm -f "$MONITOR_LOCK_FILE"
    fi
  fi
}
```

Roflicide: hm, what's `lock-monitor.pid`? Right now theres a file `flock` uses to maintain the lock called `monitor.lock`, this file is always there (that's just how flock works I believe). there's also a file called `monitor.pid` that stores the PID of the monitor.

zanix: `lock-monitor.pid` is what `createLockFile` creates

zanix: If you change `$(createLockFile $WORLD "monitor" )` to `$(createLockFile $WORLD "lock" )` the lock file would be `lock-lock.pid`

Roflicide: so i think the issue with this is that the `createLockFile` function uses `$$` as the PID for the lock file, and `$$` expands to the PID of the current shell process. for the server monitor I'm generating a new process which has a _different PID_ than the current shell process. because of this, I believe that what happens is the lock files don't track the correct PID and the whole thing breaks.

the `createLockFile` function is fine for the use cases we have now--backups and overviewer mapping-- because it's checking to make sure that the _current shell PID_ isn't performing at multiple places at once, say, trying to perform a backup while a previous backup is still underway. it's not tracking multiple processes, it's just tracking one process--the current shell process. at least, thats my current understanding, lol.

`flock`, on the other hand, lets you track/lock multiple processes, and it manages all of the PID tracking for you. 

in conclusion, because  I'm creating a new process for the server monitor I don't think using `createLockFile` will work. and my testing validates this, I think--I did do it the way you described above initially when I was creating this PR and it didn't work, and I just tested it again with the changes you suggested and unfortunately it didn't work this time either.

this begs the question of why did I spin off a new process for the server monitor in the first place? well, I initially thought I wasn't going to do that, but then I realized that since it's a monitor i think it has to be a separate process, I dont think it can be the same process as the current shell.

let me know if you have any questions, this is just my understanding of the situation and I've been completely wrong before :blush:

thanks

zanix: Makes sense, thanks for diving into it

zanix: I think the only way to make `createLockFile` work is if you rewrite it.
I would hate to introduce another dependency for the script but it seems it could be fairly common.

> The flock command is part of the util-linux package and is available from Linux Kernel Archive (https://www.kernel.org/pub/linux/utils/util-linux/).

Either way, we should still add it to the `# Required Software` section and to the docs

Roflicide: yea, i think flock is pretty common. it's also quite robust, so I'd feel comfortable using it

sandain:
What is the status of this PR? Has anyone been able to test it?  I don't have time right now (switching jobs, moving to a new state, etc).  The only change request I see is to add documentation for the new options..

Integration with my PR #274 can come before or after this PR as far as I'm concerned..

zanix:
I haven't had time to test this, and @Roflicide still need to make a couple changes.

Roflicide (commented):
i think ive made all the changes requested, just gotta update the docs and good to go!

Roflicide:
Just updated the docs! also, i learned today that `flock` is part of the `util-linux` package, which is a standard linux package shipped with all linux distros, so i think most people should already have it! but i've updated the install instructions just in case.

anyone else is welcome to do more testing with this PR if they'd like! i'd probably like to do a little more testing, but also good to merge too.

msctl
```diff
@@ -1274,6 +1278,123 @@ serverConsole() {
...
+# ---------------------------------------------------------------------------
+stopServerMonitor() {
+  local WORLD_DIR MONITOR_LOG MONITOR_PID MONITOR_LOCK_FILE ACQUIRED_LOCK
+  WORLD_DIR="$WORLDS_LOCATION/$1"
+  MONITOR_LOG="$WORLD_DIR/logs/mscs.monitor.log"
+  MONITOR_PID="$WORLD_DIR/monitor.pid"
+  MONITOR_LOCK_FILE="$WORLD_DIR/monitor.lock"
+  # Check if server monitor instance currently running.
+  (
+  flock -n 9
```
zanix, line 1303:
```suggestion
  $FLOCK -n 9
```

msctl
```diff
@@ -1274,6 +1278,123 @@ serverConsole() {
...
+  MONITOR_LOG="$WORLD_DIR/logs/mscs.monitor.log"
+  MONITOR_PID="$WORLD_DIR/monitor.pid"
+  MONITOR_LOCK_FILE="$WORLD_DIR/monitor.lock"
+  RESTART_AFTER_CRASH=$(getMSCSValue "$1" "mscs-restart-after-crash" "$DEFAULT_RESTART_AFTER_CRASH")
+
+  # Verify option is enabled.
+  if [ "$RESTART_AFTER_CRASH" = "true" ]; then
+    # Verify that there is no monitor instance currently running.
+    (
+    flock -n 9
```
zanix, line 1379:
```suggestion
    $FLOCK -n 9
```

zanix (changes requested):
Change those calls to `flock` with `$FLOCK`

Roflicide:
> Change those calls to `flock` with `$FLOCK`

thanks for the catch, fixed!

zanix (approved):

sandain:
Sorry all, I moved across the country and don't currently have my server setup for testing purposes. Has this been tested? Is it now ready to merge?

zanix:
Hey @Roflicide, this needs a rebase

msctl
sandain, line 1376: Please modify this to use the new `true_value` method:
```patch
- if [ "$RESTART_AFTER_CRASH" = "true" ]; then
+ if true_value "$RESTART_AFTER_CRASH"; then

**Review by sandain — CHANGES_REQUESTED** [review]
@Roflicide, I left a few minor comments.  Once you rebase and fix these remaining minor issues, I'll merge this PR.  Thanks for all the work.

File: msctl [hunk 12]
```diff
@@ -1274,6 +1278,123 @@ serverConsole() {
...
+      if [ -f "$MONITOR_LOG" ]; then
+        if [ "$LOG_DURATION" -gt 0 ]; then
+          find "$MONITOR_LOG" -type f -mtime +"$LOG_DURATION" -delete
+        fi
+      fi
+      # Run the server monitor.
+      # Nohup does not allow you to pass functions. However, the code below mimics nohup behavior by doing the following:
+        # Start subshell, ignore HUP signal, redirect stdin to /dev/null, redirect stdout and stderr to log file, run in background.
+      # Also store the PID of this process for later use.
+      ( trap "true" HUP ; pid=$(exec sh -c 'echo "$PPID"'); echo $pid > "$MONITOR_PID"; serverMonitor $1 ) </dev/null 2>&1 1>>"$MONITOR_LOG" &
```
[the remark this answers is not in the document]
Roflicide: Thanks for catching these, shortening them worked!

Roflicide:
Thanks for reviewing @sandain @zanix , going to re-open this as a new PR as I accidently deleted the old fork (but I still have the code)

PR #277, opened 2021-03-15
querySendPacket runtime optimisation

Sehnryr:
replacing `sleep 1` with a while loop that checks if the `query.out` is empty or not with a timeout of 1 second


sandain:
This seems like a reasonable change. It appears that `timeout` is part of the coreutils, so there are no additional dependencies -- nice.

I don't have access to my test server at the moment (in the middle of a big move and am currently living in a hotel room), so I can't test this. If we can get an ack from @zanix or @Roflicide, I would be happy to merge this.

zanix (approved):
I managed to test this just now and I'm not seeing any issues, looks alright to me

sandain:
Thanks for testing @zanix

Issue #282, opened 2021-03-29
Connect to native Forge console 


rwtallant13:
The forge console by default allows you to use the arrow keys, autocomplete commands and stores a history you can go back through for commands, you can get a similar console experience with Fabric with the mod "JLine for Dedicated Server" which i also use. 

Is it possible to connect to this console while using this control script in any way? Maybe through the use of something like screen or tmux.

sandain:
We don't use screen or tmux anymore in this script, instead piping console output to the file `console.out`, and input from the user to the file `console.in`.  I'm not sure what would be breaking autocomplete commands, but I would think it would be possible without going back to tmux/screen.

sandain:
I just ran some tests and I think this can be done.  I played around with two terminals. In the first one, I ran:
```
touch something && tail -f something
```
And in the second I ran:
```
while read -r -n1 LINE; do printf "$LINE\n" >> something; done
```

When typing in the second terminal, it was echoed in the first.  I don't have any test servers setup right now because I am in the middle of a move.  However, if you change [Line 1304](https://github.com/MinecraftServerControl/mscs/blob/main/msctl#L1304) to be the following, it might work:
```
while read -r -n1 LINE; do
```
If you can test this change and let me know how it works, I can submit a patch (unless you want to).

sandain:
On further thought, I think the patch needs to also modify the echo on [Line 1305](https://github.com/MinecraftServerControl/mscs/blob/main/msctl#L1305) to a printf.  I have submitted PR #283 to make this easier to test.

sandain:
This seems to work a lot better.  This patch boils down to the following run in two terminals:

For the first terminal:
```
touch something && tail -f something
```

And in the second:
```
while read -r -n1 CHAR; do printf "$CHAR" >> something; done
```

You can actually see the cursor moving around in the second terminal window as expected.

sandain:
This is actually much harder than I expected.  I'm working on a patch, but it is not yet fully working.  Watch PR #283 for further updates.

PR #283, opened 2021-03-31
[WIP] Fix native console for Forge servers

sandain:
For issue #282

### DIFF
--- a/msctl
+++ b/msctl
@@ -1295,14 +1295,14 @@ syncMirrorImage() {
 # @param 1 The world server to connect to.
 # ---------------------------------------------------------------------------
 serverConsole() {
-  local LINE WORLD_DIR
+  local CHAR WORLD_DIR
   # Make sure that the world's directory exists.
   WORLD_DIR="$WORLDS_LOCATION/$1"
   # Follow the console output buffer file.
   tail -f --pid=$$ $WORLD_DIR/console.out &
   # Copy user input to the console input buffer file.
-  while read LINE; do
-    echo "$LINE" >>$WORLD_DIR/console.in
+  while read -r -n1 CHAR; do
+    printf "$CHAR" >> $WORLD_DIR/console.in
   done
 }
 


msctl
```diff
@@ -1295,14 +1295,14 @@ syncMirrorImage() {
...
-  local LINE WORLD_DIR
+  local CHAR WORLD_DIR
   # Make sure that the world's directory exists.
   WORLD_DIR="$WORLDS_LOCATION/$1"
   # Follow the console output buffer file.
   tail -f --pid=$$ $WORLD_DIR/console.out &
   # Copy user input to the console input buffer file.
-  while read LINE; do
-    echo "$LINE" >>$WORLD_DIR/console.in
+  while read -r -n1 CHAR; do
```
sandain, line 1304: Well, it turns out that `-n` is not a valid option in the dash version of read.  It works great in bash though!  Back to the drawing board on this patch..

sandain (commented):
This uses a bashism.  Need to find something more generic.

msctl
sandain: I found this bit on [stackoverflow](https://stackoverflow.com/questions/8725925/how-to-read-just-a-single-character-in-shell-script):
```
read_char() {
  stty -icanon -echo
  eval "$1=\$(dd bs=1 count=1 2>/dev/null)"
  stty icanon echo
}
```

I'll try that out later.

sandain: This is a tough problem.

This is where I am at:
```patch
diff --git a/msctl b/msctl
index 2289cfb..6dd5b6a 100755
--- a/msctl
+++ b/msctl
@@ -1289,6 +1289,12 @@ syncMirrorImage() {
   fi
 }

+read_char() {
+  stty raw -echo
+  eval "$1=\$(dd bs=1 count=1 2>/dev/null)"
+  stty -raw echo
+}
+
 # ---------------------------------------------------------------------------
 # Display the server console.
 #
@@ -1301,7 +1307,18 @@ serverConsole() {
   # Follow the console output buffer file.
   tail -f --pid=$$ $WORLD_DIR/console.out &
   # Copy user input to the console input buffer file.
-  while read -r -n1 CHAR; do
+  while read_char CHAR; do
+    # Capture Ctrl+D and exit if found.
+    if [ "$CHAR" = "$(printf "\004")" ]; then
+      exit
+    fi
+    # Fix backspace character.
+    if [ "$CHAR" = "$(printf "\177")" ]; then
+      CHAR="\010\040\010"
+    fi
+    # Echo user input.
+    printf "$CHAR"
+    # Print input to console.in.
     printf "$CHAR" >> $WORLD_DIR/console.in
   done
 }
````

I've worked around an issue with the backspace character, and I now capture Ctrl+D and exit correctly.  There may be other issues though!

More work to come.

PR #287, opened 2021-04-25
Add automatic restart on crash

Roflicide:
Continuation of #271. Code should be re-based and clean now (and updated to reflect the new `true_value` function), please let me know if you find anything! :)


sandain:
Thanks for updating the PR.  Initial inspection looks good. I'll try to give this a test as soon as I can.

sandain:
This looks reasonable to me.  I'll try to dedicate some time to giving this a try.  Thanks for updating the patch.

sandain:
Sorry, I haven't spent any time testing this yet.  I think we should try to get this merged before we get the fixshellcheck branch merged to avoid having to rebase this again.  This seems pretty straight forward, but I'm not sure how to test that it is working.  Do you just `kill` the java process?  My servers never crash (but I don't have any users anymore either)...
@zanix what do you think about merging this as-is?

zanix:
> Sorry, I haven't spent any time testing this yet. I think we should try to get this merged before we get the fixshellcheck branch merged to avoid having to rebase this again. This seems pretty straight forward, but I'm not sure how to test that it is working. Do you just `kill` the java process? My servers never crash (but I don't have any users anymore either)...
> @zanix what do you think about merging this as-is?

I think you can just `kill` the process. I haven't been able to test either and I also do not have active users on a server.
This looks straightforward to me as well, it's probably safe to merge.

PR #297, opened 2021-06-27
Dynamic Systemd Units using Service Template

LanetheGreat:
Hi, big fan of MSCS but found that using the default `mscs.service` follows a "start all servers, or nothing" approach, so using Systemd's support for template units I added a template unit for MSCS that allows server admins to make individual service units for each server via `systemctl`.

With this PR, admins can disable the default `mscs.service` and make/delete units for individual minecraft servers with the following command:
`sudo systemctl [enable|disable] mscs@[server name]`

Or stop/start/restart one, or more servers if they want, using `systemctl`:
`sudo systemctl [start/restart/stop] mscs@[server 1 name] [mscs@[server 2 name]] ...`

I've tested the install and usage on my own minecraft servers and it seems to work without any issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ MSCTL := /usr/local/bin/msctl
 MSCS := /usr/local/bin/mscs
 MSCS_INIT_D := /etc/init.d/mscs
 MSCS_SERVICE := /etc/systemd/system/mscs.service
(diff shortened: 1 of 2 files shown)


zanix:
Seems good

zanix (approved):

sandain:
This looks good. Thanks for the patch @LanetheGreat

PR #308, opened 2022-02-14
suppress error when crash detection is not used

iammattcoleman:
Check if monitor.pid exists before attempting to read from it.
This prevents the following error from appearing when stopping a server:

cat /opt/mscs/worlds/example/monitor.pid: No such file or directory


sandain:
Thanks for the patch @iammattcoleman. I had noticed this issue before, but never got around to fixing it.

PR #312, opened 2022-04-12
Fix renaming a world

### LINKED ISSUE #311: Renaming worlds doesn't work
The mscs rename command doesn't properly rename the world and sets its directories. Instead, it renames the parent folder, but inside is still the old named server.

xathon:
Fixes #311 . Also, find now operates quietly and doesn't spill errors over the command line

### DIFF
--- a/msctl
+++ b/msctl
@@ -2872,9 +2872,11 @@ case "$COMMAND" in
     # Rename the world.
     printf "Renaming Minecraft world: $1"
     mkdir -p "$WORLDS_LOCATION/$2"
-    mv $WORLDS_LOCATION/$1 $WORLDS_LOCATION/$2
+    mv $WORLDS_LOCATION/$1/* $WORLDS_LOCATION/$2
+    rm -r $WORLDS_LOCATION/$1
     if [ -d $WORLDS_LOCATION/$2/$1 ]; then
-      mv $WORLDS_LOCATION/$2/$1 $WORLDS_LOCATION/$2/$2
+      mv $WORLDS_LOCATION/$2/$1/* $WORLDS_LOCATION/$2/$2
+      rm -r $WORLDS_LOCATION/$2/$1
     fi
     createWorld "$2" "$PORT" "$IP"
     printf ".\n"
--- a/msctl
+++ b/msctl
@@ -774,7 +774,7 @@ updateVersionsJSON() {
     # Make a backup copy of the version_manifest.json file.
     cp -p "$VERSIONS_JSON" "$VERSIONS_JSON.bak"
     # Delete the version_manifest.json file if it is old.
-    find "$VERSIONS_JSON" -mmin +"$VERSIONS_DURATION" -delete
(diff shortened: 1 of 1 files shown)


msctl
```diff
@@ -2872,9 +2872,11 @@ case "$COMMAND" in
     # Rename the world.
     printf "Renaming Minecraft world: $1"
     mkdir -p "$WORLDS_LOCATION/$2"
-    mv $WORLDS_LOCATION/$1 $WORLDS_LOCATION/$2
+    mv $WORLDS_LOCATION/$1/* $WORLDS_LOCATION/$2
+    rm -r $WORLDS_LOCATION/$1
     if [ -d $WORLDS_LOCATION/$2/$1 ]; then
-      mv $WORLDS_LOCATION/$2/$1 $WORLDS_LOCATION/$2/$2
+      mv $WORLDS_LOCATION/$2/$1/* $WORLDS_LOCATION/$2/$2
+      rm -r $WORLDS_LOCATION/$2/$1
```
zanix, line 2879: I do not see the difference in the rename process here
Unless creating the directory before the mv command is what is making it fail.
The 2nd command to mv the world directory inside the main world folder shouldn't have the same issue since the directory wouldn't already exist.

xathon: With the old command, a hypothetical file `$WORLDS_LOCATION/$1/file.txt` would end up in `$WORLDS_LOCATION/$2/$1/file.txt`, since mv moves the whole directory inside the new one.

See this as an example:

```
❯ mkdir old; touch old/file; mkdir -p new
❯ mv -v /tmp/old /tmp/new
renamed '/tmp/old' -> '/tmp/new/old'
❯ ls -la /tmp/new
total 12
drwxrwxr-x  3 xathon xathon 4096 Apr 16 00:30 .
drwxrwxrwt 17 root   root   4096 Apr 16 00:30 ..
drwxrwxr-x  2 xathon xathon 4096 Apr 16 00:30 old
❯ ls -la /tmp/new/old
total 8
drwxrwxr-x 2 xathon xathon 4096 Apr 16 00:30 .
drwxrwxr-x 3 xathon xathon 4096 Apr 16 00:30 ..
-rw-rw-r-- 1 xathon xathon    0 Apr 16 00:30 file
```

What my code does is moving everything inside the old directory inside the new one, then deleting the (now empty) old one.

Though, I just realized I missed creating a directory in the 2nd command...

xathon (commented):
With the newly added mkdir command that I missed, that should be working as intended.

sandain:
Thanks for the patch @xathon. This looks correct, but I won't have a chance to test it for a little while. @zanix, if you get a chance to test it before I do and are happy with the patch, feel free to click that merge button.

I wonder how long this has been broken... I never use this feature.

xathon:
No problem! And seems like it has been broken from the start in 442d548, so I guess I'm the only one who has used it :D

zanix (approved):
Oh dang, the difference is subtle but important.
I just tested this myself and the old mv commands did not work properly.
This looks good, going to merge.

PR #319, opened 2022-09-26
Notify user when an update to MSCS is available

michael2847:
Hi,

This PR adds a feature that notifies the user if an update to MSCS is available. By default, the check will occur every 7 days (let me know if you think it should be something else), but this duration can be configured by adding `mscs-update-check-duration` property to the `mscs.properties` file and setting it to a value of your choice (in days; set value less than 1 to disable update checking). 

Here's a summary of how this feature technically works: 

1. In the makefile command, it creates a link from the user's MSCS home directory to the MSCS download directory called `mscs_install_dir`. I put this line of code in the `update` portion of the `makefile` (instead of the `install` portion), so when users update to this version the link will be created.
2. Then, if the check for updates feature is enabled (by setting the `mscs-update-check-duration` to greater than 0), the `checkForMSCSUpdates` function is executed, which just `cd's` to the install directory (using the link the  `makefile` created) and does a `git fetch --dry-run`--if this has 0 output, theres no updates, otherwise there is an update available. If an update is available, it displays a link to our documentation page on how to update. 
3. Then, a temporary file is written (called` .mscs-last-update-check`) that stores the date of the last update. Once this period has elapsed, the update-check command is called again.
4. The `checkForMSCSUpdates` function can also be called manually by running `mscs check-for-mscs-updates`. 

Changes to documentation required:
- Update command reference to include `mscs check-for-mscs-updates`
- Update `mscs.properties` documentation to include `mscs-update-check-duration`
- Update manual setup instructions to instruct user to create a symlink called `mscs_install_dir` from their MSCS home to their download directory.

We could technically extend this PR to automatically doing the update as well, but it would require additional complexity and error checking (especially if the user followed the manual install); whereas with just checking for updates, regardless if the user did the `Makefile` install or the manual install, this PR should still work since both types of installs involves `git clone`ing this repository. 

I tried to test this as thoroughly as possible, but additional testing would be greatly appreciated.

Thanks,
Michael

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,8 @@ install: adduser update
 	else \
 		ln -s $(MSCS) $(MSCS_INIT_D); \
 		update-rc.d mscs defaults; \
-	fi
-	
+	fi 
+
 adduser:
 	useradd --system --user-group --create-home -K UMASK=0022 --home $(MSCS_HOME) $(MSCS_USER)
 
@@ -34,6 +34,7 @@ update:
 	@for script in $(UPDATE_D); do \
 		sh $$script; \
 	done; true;
+	ln -sf $(shell pwd) $(MSCS_HOME)/mscs_install_dir
 
 clean:
 	if which systemctl; then \
--- a/msctl
+++ b/msctl
@@ -158,6 +158,9 @@ Actions:
     Run a detailed Query on the Minecraft world server(s). Run a detailed
     query on all world servers by default.
 
+  check-for-mscs-updates
+    Checks if their is a newer version of MSCS available.
+
 Options:
 
   -c <config_file>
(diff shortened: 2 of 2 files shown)


sandain:
This assumes that the git method was used for install. I also periodically release tarballs, and this method wouldn't work for this.

Perhaps we should add a version file that the script could refer to.

zanix:
We could ping the GitHub API and see what the latest tag or release is

zanix:
https://api.github.com/repos/MinecraftServerControl/mscs/releases

Grab `tag_name` and pass it to download the specific tag
`https://github.com/MinecraftServerControl/mscs/releases/download/$tag_name/$tag_name.zip`

Or we could use the latest download link

zanix:
I just noticed there is a `tarball_url` and `zipball_url` in the API, that's probably the easiest to grab the download or even output the URL.

PR #320, opened 2022-09-28
Updated Makefile with a safety check for user

sptoothman:
Added a safety check to see if the user exists when running make install. It was dying on me because the user already existed. Now if the user exists, it reports back the user exists and skips it.


sandain:
This looks good. Thanks for the patch

PR #347, opened 2024-09-04
Adding hook scritps to start, stop and force-stop. The scripts are ex…

KingBBQ:
…ecuted, if they exists: hooks/post-start.sh - hooks/post-stop.sh

how did I get the idea? I asked ChatGPT if there is a way to have scripts automatically executed after a stop or start in MSCS, and it halucinated... as it didnt exist (yet), i added it :)

### DIFF
--- a/msctl
+++ b/msctl
@@ -1453,6 +1453,17 @@ startServerMonitor() {
 start() {
   local EULA PID SERVER_COMMAND WORLD_DIR
   WORLD_DIR="$WORLDS_LOCATION/$1"
+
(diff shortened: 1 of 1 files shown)


sandain:
Interesting idea.  Would it make sense to also add in pre-start and pre-stop hooks as well?

KingBBQ:
fair enought - then a wider range of usage is covered. added also the prescript hooks and tested it locally.
btw, i use the script to start a python script doing some analysis of the logs...

PR #356, opened 2025-10-11
Update bash completion location

### LINKED ISSUE #355: install fails on /etc/bash_completion.d/mscs
I recently tried installing mscs on an Ubuntu 22 server. I got this error:
```
install -m 0755 msctl /usr/local/bin/msctl
install -m 0755 mscs /usr/local/bin/mscs
install -m 0644 mscs.completion /etc/bash_completion.d/mscs
install: cannot create regular file '/etc/bash_completion.d/mscs': No such file or directory
make: *** [Makefile:34: update] Error 1
```
I investigated and found that my server has no `/etc/bash_completion.d` directory even though the `bash-completion` apt package is installed.
I found [this stackexchange discussion](https://unix.stackexchange.com/questions/416185/where-to-install-bash-completion-scripts-for-out-of-tree-packages) saying that `/etc/bash_completion.d/` is a legacy dir and bash completions should instead go into `/usr/share/bash-completion/completions/`.

I also looked in `/usr/share/bash-completion/completions/` on my barebones Ubuntu 22 server and found it full of completion files, so it does seems like this is where the completion file should live.

I wonder though, is this something unique to Ubuntu? or do modern Debian releases and other linux distros also use `/usr/share/bash-completion/completions/`?

zanix:
Adds a check to the bash completion location to detect where to place the completion file

Fixes #355

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,16 @@ MSCS := /usr/local/bin/mscs
 MSCS_INIT_D := /etc/init.d/mscs
 MSCS_SERVICE := /etc/systemd/system/mscs.service
 MSCS_SERVICE_TEMPLATE := /etc/systemd/system/mscs@.service
-MSCS_COMPLETION := /etc/bash_completion.d/mscs
+
+# Bash completion: detect proper location
+BASH_COMPLETION_DIR := $(shell \
+	if [ -d /usr/share/bash-completion/completions ]; then \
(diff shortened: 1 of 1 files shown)


icolwell (approved):
Tested, works!
Thanks for doing the PR :+1:

sandain:
Sorry, I didn't notice this PR. I'm slacking on my duties here I guess.

